PR #1, opened 2015-11-15
Implement multiple acronyms in !!acro and UQ parking with !parking

JosephGarrone:
Can fetch multiple acronyms

### DIFF
--- a/scripts/acronym.js
+++ b/scripts/acronym.js
@@ -9,7 +9,7 @@ var cheerio = require("cheerio");
 
 module.exports = function (robot) {
   robot.hear(/^!!acro (.+)/i, function (res) {
-    robot.getAcronyms(res, res.match[1].split(" "));
+    robot.getAcronyms(res, res.match[1].trim().split(" "));
   });
   
   robot.getAcronyms = function (res, list) {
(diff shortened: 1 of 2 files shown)


JosephGarrone:
Can now fetch the parking with !parking.

JosephGarrone:
Added some aliases to the car park names

scripts/acronym.js
```diff
@@ -7,22 +7,69 @@
...
+  // Parses the DOM to get the acroynm expansion
+  robot.getAcronym = function (html) {
+    var $ = cheerio.load(html);
+    var responses = [];
+    
+    $('td.acr').each(function (i, element) {
+      responses.push($(element).siblings().text());
+    });
+    
+    return responses[Math.floor(Math.random() * responses.length)];
```
TRManderson: I still have mixed feelings about choosing randomly, rather than taking the first

JosephGarrone: This is like mon's im pretty sure, ensure that when people acro multiple times it is never the same.

TRManderson: Choosing from the top 10 maybe?

JosephGarrone: There are generally only about 5 results anyway

PR #4, opened 2015-11-16
Added urban dictionary support

matthew-lake:


jgat:
what do you do if you want to define a phrase with more than one word?

It looks like `!urban foo bar` will look up `foo` and `bar` separately, but not `foo bar`.
Maybe it makes sense to only support one phrase per command.

scripts/urban.js
```diff
@@ -0,0 +1,60 @@
...
+							var response = "";
+							for (var j = 0; j < responses.length; j++) {
+								var word = list[j].toUpperCase();
+								var definition = robot.getUrbanDef(responses[j]);
+								var example = robot.getUrbanExample(responses[j]);
+								response += ">" + word + ": " + definition + "\r\n";
+								response += "> \t _" + example + "_ \r\n";
+							}
+
+							if (list.length > 5) {
```
TRManderson: should probs be `> limit`

matthew-lake:
good point - I'll probably change it so it supports multi-word entries but not looking up multiple words at once

scripts/urban.js
matthew-lake: I blame @JosephGarrone

scripts/urban.js
```diff
@@ -0,0 +1,60 @@
...
+
+	robot.getUrban = function (res, list) {
+		var completed = 0;
+		var responses = [];
+		var limit = 5;
+
+		var max = list.length;
+		if (list.length > limit) {
+			max = limit;
+		}
```
jgat: alternative: `max = Math.min(list.length, limit)`

PR #7, opened 2015-11-16
check for no result

matthew-lake:


TRManderson:
Please test more before PR-ing in future :+1:

matthew-lake:
:( sorry

PR #11, opened 2015-11-16
Added gImages search

BodenM:
THIS TIME IT SHOULD ACTUALLY WORK
bloody useless Github for Windows not updating for local changes.


package.json
```diff
@@ -16,6 +16,7 @@
     "cheerio": "^0.19.0",
     "hubot-google": "^0.1.2",
     "hubot-voting": "^1.0.0"
```
TRManderson: need a comma

BodenM:
fixed

external-scripts.json
```diff
@@ -4,4 +4,5 @@
   "hubot-google",
   "hubot-redis-brain",
   "hubot-voting"
```
jgat: Comma needed here as well.
It would help if you put a comma on the line below so that people editing this file in the future don't make the same mistake.

BodenM:
there :P

jgat:
Thanks :)

jgat:
Actually I probably screwed up, I don't think JSON supports trailing commas like that. I'll fix it.

PR #13, opened 2015-11-16
Shhhhh

wisebaldone:
Talks to the user in a private chat to help reduce clutter.


TRManderson:
Come on Evan, it's only one little message...

PR #14, opened 2015-11-16
I would like to add a reaction gif/image bot script thing

cavejay:
So the idea is that users can register and use images/gifs with the bot and then anyone can use the registered gifs/images.

It just means leggy could have a !mrw duck command that was a rubber duck.
Think 'bigger, free emotes'

atm the system could be abused and used to link to any resource or just repeat text. I'm going to look into enforcing links only and maybe somehow checking them to ensure validity.

tl;dr register text with bot and have the bot spit it back out as needed.

### DIFF
--- a/scripts/reaction.js
+++ b/scripts/reaction.js
@@ -19,14 +19,23 @@ module.exports = function (robot) {
     if(!url || !rStr) return;
     robot.storeReaction(rStr, url, res);
   }
-  
-  
 	
-   robot.showReaction = function(reactionString, res) {
-	   
-   }
-   
-   robot.storeReaction = function(reactionString, reactionURL, res) {
-	   
-   }
+  robot.showReaction = function(reactionString, res) {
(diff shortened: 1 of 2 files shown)


scripts/reaction.js
```diff
@@ -0,0 +1,48 @@
...
+// 
+
+module.exports = function (robot) {
+  robot.hear(/^!mrw (.+)/i, function (res) {
+	  var rStr = res.match[1];
+	  if(!rStr) return;
+	  robot.showReaction(rStr, res)
+  });
+  
+  robot.hear(/^uqcsbot new reaction (\w+\ [^ \n]+)/i, function (res) {
```
TRManderson: you can get rid of the ^uqcsbot part and just use robot.respond instead. This functions a lot better, because currently you can't PM uqcsbot things that start with ^ in the regex. Can you change that pls?

cavejay:
updated with requested change

PR #28, opened 2016-03-05
Added uqfinal.js

lewisjb:
uqcsbot <uqfinal|wdinotf|final> <score1> <score2> ... - (scores are the percentages) Returns the minimum mark on the final exam to pass (doesn't include hurdles)

### DIFF
--- a/scripts/uqfinal.js
+++ b/scripts/uqfinal.js
@@ -1,0 +1,44 @@
+// Description
+//   UQFinal
+// 
+// Commands:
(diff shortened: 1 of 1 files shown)


scripts/uqfinal.js
```diff
@@ -0,0 +1,44 @@
+// Description
+//   UQFinal
+// 
+// Commands:
+//   uqcsbot <uqfinal|wdinotf|final> <score1> <score2> ... - (scores are the percentages) Returns the minimum mark on the final exam to pass (doesn't include hurdles)
```
TRManderson: add a thing to say `<course>`

Issue #41, opened 2016-04-14
Add a killswitch


jgat:
It's possible that uqcsbot will have a bug that does annoyingly spammy things (e.g. [ping a random user every minute](https://github.com/UQComputingSociety/uqcs-hubot/commit/c9947f7225ac14193305d71dc7ab303000ce7cac)).

Suggestion: add a killswitch phrase - if someone says a particular phrase, uqcsbot will stop what it's doing (e.g. leave the channel, or exit the process it runs on, etc.).

Specific requirements:

- the kill phrase is specific and memorable enough that it won't be said by accident, and that the intention is clear. Possible examples: "uqcsbot: stopstopstop" or "uqcsbot: quitquitquit".
- there should be some indication that the action was successful (e.g. the bot says "okay" before stopping, or the bot visibly leaves the channel).

jgat:
turns out that's already a thing. :+1:
https://github.com/UQComputingSociety/uqcs-hubot/blob/master/scripts/go-down.coffee

JosephGarrone:
I'm pretty sure `uqcsbot go down` is just for restarting. Are you looking for something more like `uqcsbot leave`?

jgat:
I don't know any coffeescript, but from the looks of go-down.coffee, it just exits the process (unless heroku happens to start it up again when the process exits?)

JosephGarrone:
Yeh I think it must have a `forever` type service running. Or its daemonised. Maybe @TRManderson knows?

TRManderson:
Runs under supervisor on uqcloud. A few people have access to the Supervisor dashboard to shut it down in that situation.

PR #42, opened 2016-04-14
Add "emergency stop" documentation to README

jgat:
Makes it more obvious.

(See issue #41)


jgat:
(you don't have to accept this if you don't want, I don't care that much)

TRManderson:
Just noticed this is still here. I'm closing the PR because uqcsbot auto restarts via supervisor, so this isn't quite accurate.

PR #49, opened 2016-04-28
Added 100*n member notifications

lewisjb:


lewisjb:
No clue why it has all those old commits bc they were in #45

PR #54, opened 2016-05-25
Add a better emoticon than /shrug

gricey432:


gricey432:
Never even done node or hubot, just guessing how this works

lewisjb:
"module.exports = function (spiderShrug) {"
should be
"module.exports = function (robot) {"

If you run hubot in cmd/terminal you can test scripts and commands.

PR #57, opened 2016-05-28
Fixed spider documentation error and new voting

lewisjb:


JosephGarrone:
Looks pretty dank to me!

PR #72, opened 2016-07-03
Fix package.json

JosephGarrone:
Revert the package.json to the old order. Sorry pepes, I shouldn't have :shipit: :)


TRManderson:
No need to revert

PR #73, opened 2016-07-03
Added a passthrough system for external development

gricey432:
Please review. As always, I've never done node

### DIFF
--- a/scripts/passthrough.js
+++ b/scripts/passthrough.js
@@ -34,7 +34,6 @@ module.exports = function (robot) {
                     form: {
                         "channel_name": envelope.room,
                         "user_id": envelope.user,
-                        "user_name": envelope.user_name,
                         "text": envelope.message.text,
                     }
                 });
--- a/scripts/passthrough.js
+++ b/scripts/passthrough.js
@@ -1,0 +1,45 @@
+// Description
+//   Allows UQCSbot to pass through the messages it recieves to an unlimited number of endpoints
+//   Saving precious Slack integrations
+//
+// Commands:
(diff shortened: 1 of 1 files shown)


lewisjb:
The only thing that might be a problem is that the endpoints should be open source. Considering how people reacted with logging people might not be happy with not knowing what the endpoints are being used for,

gricey432:
@lewisjb very good point, maybe it shouldn't have the user's name in it

lewisjb:
The other thing is that currently (unless I'm mistaken) there's no way to get information back from the endpoint which **really** limits the possible functionalities

jack775544:
We could use the Incoming Webhooks slack integration to send a message back to the chat room. We actually already have it included so we don't have to worry about going over the integration limit yet

gricey432:
I've removed the user name but kept the unique user id, this should make it pretty much anonymous but still usable for identifying users

gricey432:
So, what are the remaining issues with this?

wisebaldone:
" Considering how people reacted with logging, they might not be happy with not knowing what the endpoints are being used for.". For endpoints this is a non issue since to have info "recorded" you prompt for it yourself in a command.

lewisjb:
@wisebaldone In the current `passthroughs`, it has the pattern of `/.*/`, so *every* message that uqcsbot sees, will be sent to the endpoint. But you are right in that, if the only passthroughs have command patterns it won't be an issue.

wisebaldone:
yep I assumed that was intended so it could be tested

PR #74, opened 2016-07-04
Added a way to find out your own user id

gricey432:
The passthrough thing uses user IDs rather than names, if you want to use it to track yourself, you can get your own ID and identify yourself or anyone who offers you their id

### DIFF
--- a/scripts/id.js
+++ b/scripts/id.js
@@ -1,0 +1,12 @@
+// Description
+//   Get slack user ID
+//
+// Commands:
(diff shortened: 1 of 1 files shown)


lewisjb:
Regex should be `/!?id/` instead of `/!id/` as the bot's alias is `!`, so the user would need `!!id`

gricey432:
Thanks @lewisjb , I always wondered why people were putting `?` on all the regexs

PR #91, opened 2016-08-01
Fixed voteythumbs count & new message

lewisjb:
- Voteythumbs previously required the number of up votes - down votes
   to be greater than half the channel. Now it just requires >50% up or
   >50% down


TRManderson:
lgtm at a glance

PR #95, opened 2016-08-30
Added moss cat

jack775544:
Added moss cat command to uqcsbot.
No surprise, it prints the moss cat.


jgat:
You call that a moss cat?

jack775544:
I feel like this is a war that no one can win

joeladdison:
Unfortunately the cat looks a bit worse than back in 2014 😢

PR #123, opened 2017-03-14
Update !whatsdue

mitchmcdee:
Simplified formatting further and supported no course codes being entered (to throw error for user).

### DIFF
--- a/scripts/whatsdue.js
+++ b/scripts/whatsdue.js
@@ -99,7 +99,7 @@ module.exports = function (robot) {
      * Robot responds to a message containing `!whatsdue`.
      */
     robot.respond(/!?whatsdue ?((?:[a-z]{4}[0-9]{4} ?)+)?/i, function (res) {
-        // If there is not matching groups, throw an error.
+        // If there are no matching groups, throw an error.
         if (!res.match[1]) {
             res.send('Please provide at least one valid course code.');
             return;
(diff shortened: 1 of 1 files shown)


scripts/whatsdue.js
```diff
@@ -97,9 +97,14 @@ module.exports = function (robot) {
 
     /**
      * Robot responds to a message containing `!whatsdue`.
-     * Requires at least one course.
      */
-    robot.respond(/!?whatsdue ((?:[a-z]{4}[0-9]{4} ?)+)/i, function (res) {
+    robot.respond(/!?whatsdue ?((?:[a-z]{4}[0-9]{4} ?)+)?/i, function (res) {
+        // If there are no matching groups, throw an error.
+        if (!res.match[1]) {
```
TRManderson: Is this the best check for no match?

mitchmcdee: res.match[1] is only populated by valid matches, if it doesn't exist; there are no matches. Is there another check you have in mind/can recommend?

TRManderson: `res.match.length`? I think I'm okay with it as is though

TRManderson:
Let me know when you're happy for me to merge

scripts/whatsdue.js
mitchmcdee: happy to change it if there is another compelling reason to :) but atm I think it reads better than if(res.match.length != 2) or whatever hehe

appreciate the comment Tom!

mitchmcdee:
all good to go :)

PR #125, opened 2017-03-14
Added support for querying !whatsdue in a course room to get that course's assessment

mitchmcdee:


mitchmcdee:
PTAL @TRManderson :)

PR #126, opened 2017-03-14
Updated ecp script. Fixed ecp for course room. Cleaned up some things and added error checking.

mitchmcdee:


mitchmcdee:
PTAL @TRManderson :)

PR #132, opened 2017-03-20
Fixed issue with !ecp with course codes that don't follow ABCD1234

gricey432:


TRManderson:
Do courses that don't follow that format actually exist?

gricey432:
Yes, e.g. MGTS1965D

PR #136, opened 2017-03-24
Hopefully fix race condition

somebody1234:


lewisjb:
"Hopefully", I'm assuming that means you didn't test it?

TRManderson:
What's the race condition?

lewisjb:
@TRManderson based off the code changes, I'd say the race condition of sending "Here are my commands", and the actual list of commands. I think concatenating them into 1 message would be a better solution, but that's not a big deal

TRManderson:
Yep just tested it to confirm

somebody1234:
Yeah sorry seems like the slack adapter for hubot doesn't support sending multiple messages

TRManderson:
Gonna change up the deployment via Jenkins then I'll merge this PR to test it

PR #137, opened 2017-03-26
!events: Fix recurring events

lewisjb:
Fixes recurring events in !events
- Cleaned up some old code in the process.
- The increasing ical version is due to a fix they made for their rrule stuff
- Also converted it back to JS because it was breaking peoples' local testing

Tested locally


scripts/events.js
```diff
@@ -0,0 +1,110 @@
...
+					// Repeated event, stored in the format from the rrule package
+					var times; // List of repeated times in the desired time-frame
+					if(to) {
+						times = data[k].rrule.between(new Date(now), new Date(to));
+					} else {
+						// Technically until a year ahead, safer this way.
+						times = data[k].rrule.between(new Date(now), new Date(now + week * 52));
+					}
+					var obj;
+					var duration = data[k].end.getTime() < data[k].start.getTime(); // How long it goes for
```
mitchmcdee: won't this return a boolean?

scripts/events.js
```diff
@@ -0,0 +1,110 @@
...
+					if(!to || data[k].start.getTime() < to) {
+						arr.push(data[k]);
+					}
+				}
+			}
+			arr.sort(function(a, b) {
+				return a.start.getTime() - b.start.getTime();
+			});
+
+			if(cap) {
```
mitchmcdee: why isn't this done in the loop?

mitchmcdee:
LGTM otherwise

scripts/events.js
lewisjb: You are 100% correct, this is a typo, although not a clue how I did it...

scripts/events.js
lewisjb: Well, this was here before adding recurring event support, and it is because if you just broke out of the loop when you had enough, you would have to sort before the loop. Seeing as the most common use-case only deals with a small percentage of the total number of events, iterating over them all, sorting, then truncating seemed like the nicest solution.

However, now that recurring events are also supported, this approach works even better. When you come across a recurring event, you can't know how many following events you care about (as it depends on how many other events there are etc); so adding them all, then truncating is a very clean solution in my opinion.

PR #142, opened 2017-04-02
Fixed ordering issues with !reactify

tgrkzus:
Fixes the issue, but no guarantee this code is good at all.


mitchmcdee:
I think the Promises could be implemented a little cleaner-- check out my 'whatsdue' script for some tips maybe? I'll provide a bit better feedback in a bit :)

package.json
```diff
@@ -15,15 +15,16 @@
...
     "hubot-rules": "^0.1.1",
     "hubot-scripts": "^2.16.2",
     "hubot-slack": "^4.0.0",
     "hubot-wolfram": "1.0.9",
     "hubot-youtube": "^1.0.1",
-    "hubot-postgres-brain": "^0.1.0",
     "ical": "^0.5.0",
     "kd.tree": "^0.1.3",
     "mathjs": "^3.2.1",
+    "promise": "^7.1.1",
```
mitchmcdee: I don't think this is necessary

scripts/reactify.js
```diff
@@ -45,27 +45,40 @@ var emoji = {
 	'0': ["zero"]
 };
 
+var Promise = require('promise');
```
mitchmcdee: I don't think this is necessary

scripts/reactify.js
```diff
@@ -45,27 +45,40 @@ var emoji = {
...
+
+			setTimeout(function() {
+				resolve(true);
+			}, 1000);
+		});
+	};
+
+	// We have to sequentially perform api calls for reaction otherwise the
+	// ordering can get messed up.
+	function generateReactionChain(msg, item) {
```
mitchmcdee: This implementation of Promises seems incorrect. I think a better way to do it would be create a bunch of promises and then do Promise.all(promises). For your application this would look something like:

```promises = []
msg.forEach(char => {
    var reaction = emoji[char][Math.floor(Math.random() * emoji[char].length)];
    promises.push(addReaction(null, item, reaction));
}

new Promise.all(promises)
    .then(whatever)
    .catch(whatever)```

File: scripts/reactify.js [hunk 4]
```diff
@@ -45,27 +45,40 @@ var emoji = {
...
+	
+	// Adds reaction to the provided item (a message)
+	var addReaction = function(item, emoji, callback) {
+		return new Promise((resolve, reject) => {
+			robot.adapter.client.web.reactions.add(emoji,
+			{"channel": item.room, "timestamp": item.id},
+			callback);
+
+			setTimeout(function() {
+				resolve(true);
```
mitchmcdee: ```setTimeout(() => resolve());``` would suffice (no need for 'true')

scripts/reactify.js
tgrkzus: I don't believe Promise.all(...) will work. It processes promises in parallel which doesn't solve the issue I'm aiming to solve. My original implementation of promises solves each promise before moving onto the next. I've never used promises before (or JS really) so maybe my understanding is incorrect.

mitchmcdee: Ohhhhh I see, my mistake :)

lewisjb:
Apologies for my review seeming late now. It has been "pending", I was unaware I needed to publish it.

scripts/reactify.js
```diff
@@ -45,27 +45,40 @@ var emoji = {
...
+			}, 1000);
+		});
+	};
```
tgrkzus: I just tried using the promise from the `robot.adapter.client.web.reactions.add`, still has the same ordering issues. I think timeout might be the best solution/workaround for now.

TRManderson:
@tgrkzus pls resolve conflicts

Is this good to merge once conflicts are fixed?

tgrkzus:
Hi, I think I merged correctly. A quick test seems to work fine.

PR #143, opened 2017-04-04
Implemented !vote, button-based alternative!

tgrkzus:
Fancy voting buttons. ~~There's a couple obvious issues such as people voting multiple times, having no way to undo a vote and potentially annoying highlighting when someone votes. Decided to implement this separate from !voteythumbs for these reasons.~~

In order to run this another environmental variable is needed: `SLACK_VERIFICATION_TOKEN` which is just the verification token in the app (assuming the bot becomes an app). Also an interactive messages URL needs to be set. My setup just has this as <server>/vote right now but that's easy enough to change. 

Another thing to note is if uqcsbot is in app form this could probably be implemented as /vote and be cleaner.

Demo (Abstain removes existing vote):
vote

### DIFF
--- a/scripts/button_listener.js
+++ b/scripts/button_listener.js
@@ -43,7 +43,6 @@ module.exports = function (robot) {
         // Increment vote counter
         var title = attachments[0].fields[type].title;
         var count = parseInt(title[title.length - 2]) + 1;
-        robot.logger.info(count);
         
         if (type == YES_VOTE) {
             attachments[0].fields[type].title = "Yes (" + count + ")";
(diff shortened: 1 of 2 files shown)


gricey432:
Can't say I'm keen to merge this unless people can only vote once, is there a good reason for that? Just seems half done

tgrkzus:
I've significantly improved this the exact implementation is a little dodgy and could probably be improved. I'm open to suggestions otherwise I think this is mostly functional and would be nice to test out with more than 1 person.

PR #145, opened 2017-04-10
Fixed some issues in reactify

gricey432:
Fixed issue where it could roll the same react multiple times
Fixed issue where it fails if any non-alphanumeric/space chars are used

### DIFF
--- a/scripts/reactify.js
+++ b/scripts/reactify.js
@@ -46,7 +46,7 @@ var emoji = {
 };
 
 module.exports = function (robot) {
-	robot.respond(/!?reactify ?([a-zA-Z0-9 ]+)?$/i, function (res) {
+	robot.respond(/!?reactify ?(.+)?$/i, function (res) {
 		if (robot.adapter.client && robot.adapter.client.rtm) {
 			// Adds reaction to the provided item (a message)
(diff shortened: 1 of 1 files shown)


scripts/reactify.js
```diff
@@ -56,16 +56,22 @@ module.exports = function (robot) {
...
+			// Get text, all in lowercase and remove non alphanumeric
+			var msg = res.match[1].toLowerCase().replace(/[^a-zA-Z0-9]/g, "");
 
 			// React in reverse order since it seems reacts occur in
 			// first in last out order.
+			var used_reacts = [];
 			for (var i = msg.length - 1; i >= 0; i--) {
 				// Get random emoji from our dictionary of lists
-				var react = emoji[msg[i]][Math.floor(Math.random() * emoji[msg[i]].length)];
+				var possible_reacts = emoji[msg[i]].filter(x => used_reacts.indexOf(x) < 0);
```
TRManderson: not 100% sure this is supported

TRManderson: I just double-checked, should be OK

gricey432: Yeah that's why I was asking about ES6 support

PR #147, opened 2017-05-05
Added !ping

tgrkzus:
Bot has to have admin permissions. `!ping (everyone|here|channel)` Will only work for users in the allowed list, or in channels with users less than 50 users. I added all the committee to the allowed list. Not tested very well.

### DIFF
--- a/scripts/ping.js
+++ b/scripts/ping.js
@@ -1,0 +1,43 @@
+// Description
+//   Restricts channel and everyone pinging to predefined people
+//   Stops people from pinging everyone out of ignorance (mostly)
+//   This command works for anyone in channels with less that 50 people
+//   
+// Commands:
+//   !`ping` (channel|everyone|here)
+
+// Dictionary of characters mapping to emojis.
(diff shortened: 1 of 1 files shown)


TRManderson:
Looks good to me.

lewisjb:
Would be nice if there was an error message if it wasn't allowed / unknown ping type.
Seeing as it requires uqcsbot to be an admin, caesar.js will also need to be changed (as you could use it to do pings in larger channels).

TRManderson:
A full-on per-channel permission system is probably a bit much to ask, but it'd be cool

scripts/ping.js
```diff
@@ -0,0 +1,44 @@
+// Description
+//   Restricts channel and everyone pinging to predefined people
+//   Stops people from pinging everyone out of ignorance (mostly)
+//   This command works for anyone in channels with less that 50 people
+//   
+// Commands:
+//   !`ping` (channel|everyone|here)
+
+// Dictionary of characters mapping to emojis.
```
jgat: Leftover comment from a copy-paste.

PR #154, opened 2017-08-18
Importable calendars for assessment

mitchmcdee:
Added !calendar script which piggy-backs off !whatsdue to collate assessment pieces and generate an importable calendar containing them.

Few things I still want to clean up but would love some code review / testing :)


mitchmcdee:
Oh this PR also cleans up some of the whatsdue code and gives some better formatting

lewisjb:
Looks good so far! Should we replace all usages of the `ical` library with `ical.js`? (Seems odd having two libraries that do the same thing)

mitchmcdee:
Tidied up some stoof, I think it's ready for push unless someone points out any additions :)

RE: two ical libraries, I agree-- I just gave it 5 seconds effort to remove 'ical' but couldn't immediately find a suitable function for 'fromUrl' so I think it'll take a bit of conjiguring lel. Maybe we should make another issue for deprecating 'ical' ?

TRManderson:
LGTM, but going forward I'd like to make use of the Hubot internal webserver to create a URL for calendar download instead of doing an upload.

mitchmcdee:
when we port to python eventually that sounds like a good move (Y)

Issue #116, opened 2016-11-12
!weather radar not always detected


lewisjb:
Due to how weather.js works, sometimes using "!weather radar" will activate the normal "!weather" call.

Suggested fixes:
 - Only have one `robot.respond`
 - Make the RegEx stricter

mitchmcdee:
could you give some examples of what inputs trigger this?

lewisjb:
`!weather radar` & `!weather radar brisbane`, neither run the radar code.

gricey432:
Is this still needed now that we have `!radar`?

mitchmcdee:
I think if we're going to close this ticket and accept having a command for each functionality, we should remove radar code from !weather first :)

mitchmcdee:
Fixed in 33c14a3, closing :)

PR #155, opened 2017-08-20
Add a bom radar

gricey432:


gricey432:
Kind of related to #116 but not really

scripts/radar.js
```diff
@@ -0,0 +1,11 @@
+// Description
+//   Gets the Beureau of Meteorology's latest radar image
+// Commands
+//   !`radar` - Gets Brisbane's radar
+//
+
+module.exports = function (robot) {
+	robot.respond(/!?radar/i, function (res) {
+    res.send("https://bom.lambda.tools/?location=brisbane&_cache=" + Date.now());
```
CameronAavik: I think there should be a comment here linking to the docs for this in case it breaks and we don't know why

scripts/radar.js
```diff
@@ -0,0 +1,11 @@
+// Description
+//   Gets the Beureau of Meteorology's latest radar image
```
CameronAavik: Bureau*

scripts/radar.js
```diff
@@ -0,0 +1,11 @@
...
+// Commands
+//   !`radar` - Gets Brisbane's radar
+//
+
+module.exports = function (robot) {
+	robot.respond(/!?radar/i, function (res) {
```
CameronAavik: and remove this pesky tab :P

scripts/radar.js
gricey432: What docs?

PR #166, opened 2017-10-17
Refactored !welcome and fixed(?) member milestones

mitchmcdee:
As per title :)

Fixes:
- Milestones (#150)
- Display names (#161)

PR #168, opened 2017-10-19
Patched !welcome script (default name and API pagination)

mitchmcdee:
As per title, fixes two bugs from #166

1) display_name is optional, so instead we now default to `display_name` and resort to `name` as a back up (which is a required field)
2) Slack Web API default limits `conversations.members` to returning only 100 members. Added pagination code that utilises `next_cursor` to accurately yield _all_ members

### DIFF
--- a/scripts/welcome.js
+++ b/scripts/welcome.js
@@ -1,9 +1,10 @@
 // Description
 //   Welcomes new users to UQCS Slack and check for member milestones
 
-var MEMBER_MILESTONE = 50;
-var MESSAGE_PAUSE = 2500;
-var WELCOME_MESSAGES = [
+var API_LIMIT = 200;        // Number of members to get at a time
+var MEMBER_MILESTONE = 50;  // Number of members between posting a celebration
+var MESSAGE_PAUSE = 2500;   // Number of seconds between sending bot messages
+var WELCOME_MESSAGES = [    // Welcome messages sent to new members
     "Hey there! Welcome to the UQCS slack!",
     "This is the first time I've seen you, so you're probably new here",
     "I'm UQCSbot, your friendly (open source) robot helper",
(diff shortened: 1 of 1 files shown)


mitchmcdee:
I just realised this code "fixes" that it will only fire when a user enters #announcements instead of #general. Should I revert it back to firing on #general?

ashfordneil:
The getNumMembers function should probably use the `user` argument that was passed to it rather than relying on `announcements` being defined in a previous scope. Also is there any chance that the slack api offers a way for us to get the number of members in one go without iterating through them one by one? 

Other than those things, assuming you’ve tested it, it looks good to me. (Unfortunately I can’t test it myself at the moment)

mitchmcdee:
Interesting, any reason why you'd argue to use `user` over `announcements`? (or did you make a mistake, because user isn't passed to getNumMembers?)

re: slack API, unfortunately it seems like no-- not that I've seen at least. The only way seems to be getting a members list, of which only pagination-type methods exist as per later this year (https://api.slack.com/changelog/2017-10-members-array-truncating). On that note though, It seems I can simplify the code because `conversations.members` doesn't seem to include deleted/deactivated users, something we had to take into account when using `users.list`

Fair about testing, it's a bit of a pain to setup anyway bahahhaa

scripts/welcome.js
```diff
@@ -13,44 +14,48 @@ var WELCOME_MESSAGES = [
...
-        });
+        // Get member count to see if we've hit a member milestone
+        getMemberCount(robot, announcements, 0).then(memberCount => {
+            // If we're not at a member milestone, don't bother celebrating!
+            if (memberCount % MEMBER_MILESTONE != 0) {
+                return;
+            }
+
+            res.send(":tada: " + memberCount + " members! :tada:");
+        }).catch(err => console.log(err));
```
CameronAavik: can eta-reduce this to `.catch(console.log)` :P

mitchmcdee: good catch :^) done

mitchmcdee:
I'm going to wait for Neil's response/approval before merging :^)

PR #169, opened 2017-10-21
Tiny wording change to no events

gricey432:


gricey432:
Discussion in slack, this is correct as is

PR #171, opened 2017-10-23
Added !wavie script

mitchmcdee:
Waves at people as they enter or leave rooms :)

### DIFF
--- a/scripts/wavie.js
+++ b/scripts/wavie.js
@@ -1,0 +1,32 @@
+// Description
+//   Wave people as they enter and exit the room
+
+module.exports = function (robot) {
+    // Wave at the latest message in the room
+    function wave(room) {
+        // Make sure we have access to all the clients we need
+        if (!robot.adapter.client || !robot.adapter.client.web) {
+            return;
+        }
+
+        // Shorten web client so we can have nice concise lines :)
+        webClient = robot.adapter.client.web;
+
+        // Try channel as a public channel search first, then as a private channel
+        webClient.channels.info(room)
+            .then(res => webClient.reactions.add('wave', {channel: room, timestamp: res.channel.latest.ts}))
+            .catch(err => { if (err == 'channel_not_found') return webClient.groups.info(room); })
+            .then(res => webClient.reactions.add('wave', {channel: room, timestamp: res.channel.latest.ts}))
+            .catch(err => console.log);
+    }
+
+    // Listen for when someone enters a room
+    robot.enter(function (res) {
+        wave(res.message.room);
+    });
+
+    // Listen for when someone leaves a room
+    robot.leave(function (res) {
+        wave(res.message.room);
+    });
+};


scripts/wavie.js
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        // Shorten web client so we can have nice concise lines :)
+        webClient = robot.adapter.client.web;
+
+        // Try channel as a public channel search first, then as a private channel
+        webClient.channels.info(room)
+            .then(res => webClient.reactions.add('wave', {channel: room, timestamp: res.channel.latest.ts}))
+            .catch(err => { if (err == 'channel_not_found') return webClient.groups.info(room); })
+            .then(res => webClient.reactions.add('wave', {channel: room, timestamp: res.channel.latest.ts}))
```
TRManderson: By the time we've had a failure and re-requested, there's a reasonable chance there'll be an extra message in the channel. We should be able to get the original message timestamp from the argument to the `robot.enter`/`robot.leave` callback.

scripts/wavie.js
```diff
@@ -0,0 +1,32 @@
...
+        // Make sure we have access to all the clients we need
+        if (!robot.adapter.client || !robot.adapter.client.web) {
+            return;
+        }
+
+        // Shorten web client so we can have nice concise lines :)
+        webClient = robot.adapter.client.web;
+
+        // Try channel as a public channel search first, then as a private channel
+        webClient.channels.info(room)
```
TRManderson: We should be able to get the room from the original thing passed into the `robot.enter`/`robot.leave` callback.

mitchmcdee: I don't understand-- that's exactly what i'm passing to wave() ?

scripts/wavie.js
mitchmcdee: yeah I had that code initially but decided it overcomplicated things for (what I thought) was pretty edge case, but I think you're right.

Unfortunately, the result passed to `robot.enter`/`robot.leave` doesn't actually include a TextMessage that I can dig into (see https://github.com/hubotio/hubot/blob/master/src/message.js#L55). I was going to get this information using the `latest` attribute returned from `channels.info` and `groups.info` to check if the sub type is one of a user entering/leaving before waving, but this would mean no :wave: would occur in an "extra message" situation. Another option is something like `channels.history`, but I believe this requires additional bot privileges

TRManderson:
If 172 depended on it and it got merged then this also gets merged

scripts/wavie.js
TRManderson: Apparently I'm bad with words atm - I meant we should probably be able to determine if it's public or private from there

mitchmcdee: how so? I believe we're just given a room id-- is it determinable by the prefix letter or sum shit?

mitchmcdee:
rip

mitchmcdee:
100% happy to keep responding to your comments here and submit another PR with the fixes

scripts/wavie.js
TRManderson: Good enough

scripts/wavie.js
TRManderson: Eh, if not that's why I said "probably" - should be able to do it nicer somehow but happy to leave it as is

mitchmcdee: I investigated and room id indeed has a prefix letter that is used for distinguishing. 'G' is for groups (private), 'C' is for channels (public) and 'D' is for direct messages. I'll do a prefix check :)

PR #174, opened 2017-10-23
Add :loading: react to wakie wakie

mitchmcdee:
When wakie wakie is run, have uqcsbot hit 'em up with the ol :loading:


gricey432:
lgtm

PR #176, opened 2017-10-24
Add test slack info to readme 

tgrkzus:
I've setup a test slack with 4 bot tokens for people to use. If someone else could test they can follow my instructions and get a bot connected that'd be appreciated :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,25 @@ Then you can interact with uqcsbot by typing `uqcsbot help`.
 2. `dos2unix bin/hubot`
 3. `bin/hubot`
 
+### Running the bot on our test slack
+
+We've created a test slack just for testing bot features out.
+
+How to use:
+1. [Join the test slack by clicking here](https://uqcstest-inviter.herokuapp.com/)
+2. Pick one of the bot tokens to connect your bot to the bot on the test slack
+3. Setup a new enviromental variable: `HUBOT_SLACK_TOKEN=xoxb-YOUR-TOKEN-HERE`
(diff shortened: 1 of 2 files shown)


mitchmcdee:
Awesome job I'll check it out in a bit :^)

mitchmcdee:
Yep works for me, awesome :)

Small issue I saw though, gamma's token gave me delta as a bot-- maybe double check which is what :^)

mitchmcdee:
Also, could this be baked into the run script somehow? i.e. `./bin/hubot --adapter slack -[1-4]` to give inject the tokens for you? Or is there someway to retrieve an "unused" bot by checking exit codes or something?

tgrkzus:
Fixed the token issue.

I could probably work out a way to add it to the run script actually, I'll look into it.

tgrkzus:
I've altered the bin/hubot script to automatically set the token env variable when provided with a number 0-3. Doesn't check if multiple people are using it so it might just explode if so.

mitchmcdee:
Awesome :) I still think this can be improved EVEN MORE with totally automated bot getting

you can probably utilise https://api.slack.com/methods/users.getPresence to check if a bot is active, and return the first bot you can find that's not active else throw an error?

Besides being slick af, I think this would be even friendlier for new-comers

Again, I doubt we're going to have multiple people using bots such that collisions will be a huge issue-- so if you cbf doing this suggestion I understand :) It probably wouldn't be too hard though, you could even offload it to a python script for nicer handling

tgrkzus:
I cbf. I'll do it in a later PR (maybe for hacktoberfest :))

PR #177, opened 2017-10-24
Added list of waiting emojis for wakie.js

mitchmcdee:
Turns out :loading: isn't even an emoji, what a lol

-Added a list of waiting emojis that will be selected from at random to react with

### DIFF
--- a/scripts/wakie.js
+++ b/scripts/wakie.js
@@ -3,9 +3,10 @@
 
 var HubotCron = require('hubot-cronjob');
 
-var API_LIMIT = 200;                 // Number of members to get at a time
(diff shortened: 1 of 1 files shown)


gricey432:
Not 100% on this one. Afaik people attach the waiting after a couple of minutes. Having it just be there by default might take away the point

TRManderson:
Nah, people generally chuck it on straight away

mitchmcdee:
As tommo said, I mainly observe people chucking it on straight away, and I think doing so encourages the "victims" to respond by showing that people are interested in hearing what they're up to. Additionally, I have observed that when no one reacts to the ping, its usually forgotten about. So that's why I've created this script, to get the ball rolling and encourage interaction :^)

PR #179, opened 2017-10-24
Changed no-events message to something completely different

gricey432:
End some arguments


TRManderson:
I thought we were gonna make it "No events beep boop beep boop"?

PR #180, opened 2017-10-24
Removed broken radar code from !weather

gricey432:
#116 comment by @mitchmcdee


gricey432:
Good catch @mitchmcdee

Issue #181, opened 2017-10-25
Bug: messages containing the word "latex" trigger the latex command incorrectly 


CameronAavik:
It seems to only trigger if there are words surrounding "latex"

CameronAavik:
Looks like the issue is the [regex](https://github.com/UQComputingSociety/uqcsbot/blob/master/scripts/latex.js#L10)

    /\$\$([^\$]+)\$\$ | !?(tex|latex) ([^\$]+)/

It should be

    /\$\$([^\$]+)\$\$|!(tex|latex) ([^\$]+)/

mitchmcdee:
Fixed in 1d90a3d, closing

PR #185, opened 2017-10-25
Fix latex script

tgrkzus:
Latex script was being called at the wrong times. Fixed prob

Solves: #181


mitchmcdee:
I don't like the added boilerplate, is there anyway this can be done in one `robot.hear` ?

tgrkzus:
I've just removed the !latex alias, since no one uses it and I doubt anyone would.

PR #187, opened 2017-10-25
Created !stats for general slack analytics (anonymous information, non-invasive)

mitchmcdee:
!stats is an extensible stat tracking framework for general Slack data.

Currently tracks two statistics:

1) Number of messages in a public channel (and by extension, total messages)
2) Number of calls made on a command (and by extension, total calls)

Both statistics are non-invasive and do not store any user information at all :^)


mitchmcdee:
Example of `!stats commands`:

image

and of `!stats rooms`:

image

scripts/stats.js
```diff
@@ -0,0 +1,142 @@
...
+
+//////////////////////
+// HANDLER COMMANDS //
+//////////////////////
+
+// Handles room stat
+function handleRoomStat(stats, res) {
+    // If room was not a public channel and not a private channel, exit
+    var room = res.message.room;
+    if (room[0] != 'C' && room[0] != 'G') return;
```
CameronAavik: Could you explain what this line does? I'm a little confused.

scripts/stats.js
```diff
@@ -0,0 +1,143 @@
...
+    var command = res.message.text;
+    if (command[0] != '!') {
+        return;
+    }
+
+    incrementCounter(stats.commands, command);
+
+    // If command did not contain any options, exit
+    var baseCommand = command.split(' ')[0];
+    if (baseCommand == command) {
```
CameronAavik: could you do `if(command.indexOf(' ') < 0)` instead? (and same with other places which you check in a same way.

scripts/stats.js
mitchmcdee: room ids have letter prefixes which denote whether they are of type Channel (aka public channels, 'C'), Group (aka private channels, 'G') or Direct (aka personal messages, prefix 'D'). I check this first character to basically check the room being passed is a valid public/private channel

CameronAavik: Are you sure we want private room statistics being included?

scripts/stats.js
mitchmcdee: In the instance you mentioned I actually use the split to get baseCommand, so its necessary-- but I'll apply that change to the other location :)

scripts/stats.js
mitchmcdee: yeah that's a valid question. I came to the conclusion it was fine as it's only tracking message count, which seems pretty non-intrusive to me

mitchmcdee: If people have a problem with it though, happy to just limit it to public channels

mitchmcdee: In fact I think that might be a good idea. Private channels aren't really of interest as a statistic anyway beyond the total number of private channel messages, which is a statistic we're already given

mitchmcdee: Removed in latest commit

CameronAavik: Yeah, there's also the fact that some private channels would like to remain a secret

mitchmcdee: ah good point I thought they still showed up in list of channels

mitchmcdee:
OOoooooosk those are both great points @lewisjb. I think I underestimated just how much spam will be generated by storing all command permutations (just wanted dat extra data!), and great point about !voteythumbs.

I think the best short term solution would be to only support base commands, and decide on how to extend it to all commands in future.

As for general channel spam, I think I'll take a leaf from `!help` and send the results to the user :)

Do these changes sound good to you? Or any further comments?

lewisjb:
Yeah, they both sound good; it might also be worthwhile ignoring group chats in the command stats aswell (seeing as it's already being done in chat stats)

PR #192, opened 2017-10-30
added a subscriber system to !stats

### LINKED ISSUE #199: !stats should say what the time period is
!stats just gives a bunch of numbers with no time period, which doesn't mean much. It should say "since..."

mitchmcdee:
This PR does a few things

- Subscription system allowing people to subscribe to stats
- Stats reset and subscriptions published every Monday (might change this to fortnightly?)
- Arbitrary stats can be subscribed to and retrieved (aka `!stats !cat` will return the value associated with the `!cat` stat)
- Adds percentage statistic to room and command printouts
- Moved from API calls to RTM datastore for getting room names to remove the ability to DOS uqcsbot lol
- fixes #199 
- Removes tracking of direct messages to uqcsbot

### DIFF
--- a/scripts/stats.js
+++ b/scripts/stats.js
@@ -3,7 +3,7 @@
 //
 // Commands:
 //   `!stats (rooms|commands|<STAT>)` - Yields general slack statistics for analysis and insight
-//   `!stats (subscribe|unsubscribe) (rooms|commands|<STAT>)` - Subscribes/Unsubscribes user from the given stats
+//   `!stats (subscribe|unsubscribe) (rooms|commands|<STAT>)` - Subscribes/Unsubscribes user from the given stat
 
 var HubotCron = require('hubot-cronjob');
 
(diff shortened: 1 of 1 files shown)


mitchmcdee:
fixed @gricey432 ptal :^)

CameronAavik:
After my discussion with mitch, I think we should also make it so that the bot does not track commands that are used in non-public channels.

scripts/stats.js
```diff
@@ -92,18 +92,13 @@ function handleRoomStat(stats, res) {
 
 // Handles command stat 
 function handleCommandStat(stats, res) {
-    // If the command is not actually a command or the room is a private channel, exit
+    // If the room is not a public channel, exit
     var command = res.message.text;
     var room = res.message.room;
-    if ((command[0] != '!' && room[0] != 'D') || room[0] == 'G') {
```
CameronAavik: don't we need to keep the check that the command starts with an exclamation mark?

PR #195, opened 2017-10-31
Made !uqfinal dynamic by semester, added !dog (like !cat) and !cookbook (link)

nickl93:
All the feedback plz. ❤️

### DIFF
--- a/scripts/cookbook.js
+++ b/scripts/cookbook.js
@@ -6,6 +6,6 @@
 
 module.exports = function (robot) {
   robot.respond(/!?cookbook/i, function (res) {
-    res.send("https://jenkins.uqcs.org.au/job/cookbook/lastSuccessfulBuild/artifact/main.pdf");
+    res.send("https://github.com/UQComputingSociety/cookbook");
   });
 };
--- a/scripts/cookbook.js
+++ b/scripts/cookbook.js
@@ -1,0 +1,11 @@
+// Description
+//   Returns a link to the UQCS student-compiled cookbook (full of tasty treats).
+// Commands
+//   `!cookbook` - returns the URL of the UQCS student-compiled cookbook (pdf).
+//
+
+module.exports = function (robot) {
+  robot.respond(/!?cookbook/i, function (res) {
+    res.send("https://jenkins.uqcs.org.au/job/cookbook/lastSuccessfulBuild/artifact/main.pdf");
+  });
+};
--- a/scripts/dog.js
+++ b/scripts/dog.js
@@ -1,0 +1,19 @@
+// Description
+//   Like !cat, but for dog people (and with less trauma for 2310 students).
+//
+// Commands:
+//   `!dog` - Like !cat, but for dog people.
+
+module.exports = function (robot) {
+	robot.respond(/!?dog/i, function (res) {
+		res.send("```\n" +
+            "                                __\n" +
+            "         ,                    ,\" .\`--o\n" +
+            "        ((                   (  | __,\'\n" +
+            "         \\\\~----------------' \\_;/\n" +
+            "         (                      /\n" +
+            "         /) ._______________.  )\n" +
+            "        (( (               (( (         hjw\n" +
+            "         \`\`-\'               \`\`-\'\n\n```");
+  });
+};
--- a/scripts/uqfinal.js
+++ b/scripts/uqfinal.js
@@ -8,58 +8,92 @@
 var math = require("mathjs");
 
 module.exports = function (robot) {
-	robot.respond(/!?uqfinal ([a-z]{4}[0-9]{4}[a-z]?) (.+)+/i, function (res) {
-		//res.match[2] is the course code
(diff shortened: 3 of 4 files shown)


mitchmcdee:
it's no biggie, but in future you should probably split this up into 3 PRs so that we can seperate the features and give feedback/approval accordingly :)

scripts/dog.js
```diff
@@ -0,0 +1,19 @@
+// Description
+//   Like !cat, but for dog people (and with less trauma for 2310 students
```
mitchmcdee: forgot the ')' at the end :^)

mitchmcdee: also, could you provide a screenshot of the dog printout? seeing as though its a visual script, it'd be nice to see what it looks like for whatever reason :)

scripts/uqfinal.js
```diff
@@ -8,58 +8,90 @@
...
+          res.send("Failure fetching current semester (HTTP Error: " + resp.statusCode + ")");
+          return;
+        }
+
+        var semesters = JSON.parse(body).data.semesters;
+        for (var i = 0; i < semesters.length; i++) {
+          if (semesters[i].isCurrent) {
+            semester = semesters[i].uqId;
+            break;
+          } else {
```
mitchmcdee: dangling else statement :)

scripts/uqfinal.js
```diff
@@ -8,58 +8,90 @@
...
-						var result = Math.ceil(needed/course.assessment[num_of_assessment - 1].weight * 100);
-						res.send("You need to achieve at least " + result + "% on the final exam.\n_Disclaimer: this does not take hurdles into account_\n_Powered by http://uqfinal.com_");
-					}
-				});
-	});
-}
+  robot.respond(/!?uqfinal ([a-z]{4}[0-9]{4}[a-z]?) (.+)+/i, function (res) {
+    //res.match[1] is the course code
+    //res.match[2] is the space-seperated scores
+    var semester;
```
mitchmcdee: does this need to be hiding out here? I (personally) think it would give context to the `if (!semester)` code below to have it there

scripts/dog.js
nickl93: screen shot 2017-11-01 at 8 04 54 pm

scripts/uqfinal.js
```diff
@@ -8,58 +8,98 @@
...
+        }
+
+        // Handle failed HTTP request (check API endpoints at uqfinal.com if consistently failing)
+        if (resp.statusCode != 200) {
+          res.send("Failure fetching current semester (HTTP Error: " + resp.statusCode + ")");
+          return;
+        }
+
+        // Get the current semester uqId (as per UQ Rota)
+        var semesters = JSON.parse(body).data.semesters;
```
mitchmcdee: should probably double check the API/gricey, but on quick inspection it looks like its returning a sorted list of semesters-- so the current semester will always be the last-- so why don't we just do `JSON.parse(body).data.semesters.pop()` or something?

package.json
```diff
@@ -18,6 +18,7 @@
     "ical": "^0.5.0",
     "ical.js": "^1.2.2",
     "mathjs": "^3.2.1",
+    "purescript": "^0.11.6",
```
mitchmcdee: why was this added?

nickl93: My bad, that was me trying to get the purescript issues to go away (until trm provided a fix). Accidentally deleting package.json was me (poorly) trying to stop this from being included.

mitchmcdee: bahaha no wozzas

scripts/uqfinal.js
```diff
@@ -8,58 +8,92 @@
...
+        }
+
+        // Handle failed HTTP request (check API endpoints at uqfinal.com if consistently failing)
+        if (resp.statusCode != 200) {
+          res.send("Failure fetching current semester (HTTP Error: " + resp.statusCode + ")");
+          return;
+        }
+
+        // Get the current semester uqId (as per UQ Rota)
+        var semester = JSON.parse(body).data.semesters.pop().uqId;
```
mitchmcdee: this will error if `pop()` returns undefined for whatever reason. I'd just put the `uqId` in the `robot.http()` request below :)

scripts/cookbook.js
```diff
@@ -0,0 +1,11 @@
+// Description
+//   Returns a link to the UQCS student-compiled cookbook (full of tasty treats).
+// Commands
+//   `!cookbook` - returns the URL of the UQCS student-compiled cookbook (pdf).
+//
+
+module.exports = function (robot) {
+  robot.respond(/!?cookbook/i, function (res) {
+    res.send("https://jenkins.uqcs.org.au/job/cookbook/lastSuccessfulBuild/artifact/main.pdf");
```
TRManderson: Can you make it link to https://github.com/UQComputingSociety/cookbook instead? there's a link to the compiled PDF there but it has all the extra info too, so it's probs a better option.

PR #196, opened 2017-11-01
Culled all the shit from README

mitchmcdee:
It always felt super intimidating to me as a newbie looking at that monolithic hunk of text when getting started, so I think it's worth culling it down to just the relevant good stuff :)

Lemme know if I've accidentally removed something important lel


README.md
```diff
@@ -1,16 +1,17 @@
...
+2. `dos2unix bin/hubot`
+3. `bin/hubot`
 
 ### Running uqcsbot locally
 
-**If you're having trouble building/running the bot don't be afraid to ask for help in the UQCS slack**
+**If you're having trouble building/running the bot, don't be afraid to ask for help in the UQCS slack**
 
 You can test your hubot by running the following, however some plugins will not
 behave as expected unless the [environment variables](#configuration) they rely
```
nickl93: This link will be broken when the configuration section below is removed.

nickl93: This link will be broken when the configuration section below is removed.

mitchmcdee: good catch :^) fixed

gricey432:
Some of this longer information could be important. Instead of being removed, have you considered making a `/docs` folder and moving advanced documentation into multiple markdown files?

TRManderson:
Pretty much none of the extra info is important and it all comes from the autogenerated readme.

PR #200, opened 2017-11-06
After fighting with git to do the thing, we have an !urban

nickl93:
Have tested to confirm still works and updated instructions to post-Mitch style.

### DIFF
--- a/scripts/urban.js
+++ b/scripts/urban.js
@@ -16,36 +16,32 @@ module.exports = function (robot) {
           return res.send("> Usage: `!urban <SEARCH_PHRASE>`");
         }
 
-        // Check for JS/hubot errors
+        // Check that a response was received.
         if (err) {
           return res.send(">>> Error: " + err.toString() + ".");
         }
 
-        // Check for HTTP Errors
+        // Check for HTTP Errors.
         if (resp.statusCode !== 200) {
           return res.send("> HTTP Error ( " + resp.statusCode + ").");
         }
 
-        var udResp = JSON.parse(body);
+        var udResp = JSON.parse(body); // The urban dictionary response, parsed into a JSON object.
 
         // Parse Urban Dictionary response and send result.
-        if (udResp["result_type"] !== undefined && udResp["result_type"] !== "no_results") {
-          try {
-            var firstResult = udResp["list"][0];
-            var definition = firstResult["definition"];
-            var example = firstResult["example"];
-            var response = phrase.toUpperCase() + ":\n" + definition.toString() + "\n";
-            if (example) {
-              response += ">>> " + example.toString();
-            }
-            res.send(response);
-            if (udResp["list"].length > 1) {
-              res.send(" - more definitions at http://www.urbandictionary.com/define.php?term=" + encodeURI(phrase));
-            }
-            return;
-          } catch(err) {
-            return res.send(">>> Error parsing Urban Dictionary response: " + err.toString());
+        if (udResp["result_type"] === "exact") {
+          var firstResult = udResp["list"][0];
+          var definition = firstResult["definition"];
+          var example = firstResult["example"];
+          var response = phrase.toUpperCase() + ":\n" + definition.toString() + "\n";
+          if (example) {
+            response += ">>> " + example.toString();
           }
+          res.send(response);
+          if (udResp["list"].length > 1) {
+            res.send(" - more definitions at http://www.urbandictionary.com/define.php?term=" + encodeURI(phrase));
+          }
+          return;
         } else {
           return res.send("> No results found for " + phrase +". ¯\\_(ツ)_/¯")
         }
--- a/scripts/urban.js
+++ b/scripts/urban.js
@@ -6,68 +6,49 @@
 //
 
 module.exports = function (robot) {
-	robot.respond(/!?urban (.+)/i, function (res) {
-		robot.getUrban(res, res.match[1].trim());
-	});
+  robot.respond(/!?urban(.*)/i, function (res) {
+    var phrase = res.match[1].trim();
+    var urban = robot.http("http://api.urbandictionary.com/v0/define?term=" + encodeURI(phrase))
+      .get()(function (err, resp, body) {
 
-	robot.getUrban = function (res, word) {
+        // Check for correct usage
+        if (res.match[1][0] !== " " || !phrase || phrase.length < 1) {
+          return res.send("> Usage: `!urban <SEARCH_PHRASE>`");
+        }
 
-		robot.http("http://api.urbandictionary.com/v0/define?term=" + word).get()
-		(function (err, resp, body) {
-			if (!err) {
-				var response = "";
+        // Check for JS/hubot errors
+        if (err) {
+          return res.send(">>> Error: " + err.toString() + ".");
+        }
 
-				var max = 2;
-				var lines = 0;
+        // Check for HTTP Errors
+        if (resp.statusCode !== 200) {
+          return res.send("> HTTP Error ( " + resp.statusCode + ").");
+        }
 
-				if (robot.getUrbanResult(body) == "no_results") {
-					response += "> :( No result found for " + word;
-				} else {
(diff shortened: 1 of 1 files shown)


scripts/urban.js
```diff
@@ -0,0 +1,73 @@
...
+//
+
+module.exports = function (robot) {
+	robot.respond(/!?urban (.+)/i, function (res) {
+		robot.getUrban(res, res.match[1].trim());
+	});
+
+	robot.getUrban = function (res, word) {
+
+		robot.http("http://api.urbandictionary.com/v0/define?term=" + word).get()
```
CameronAavik: Make sure to encode the word

scripts/urban.js
```diff
@@ -6,68 +6,49 @@
...
+        if (err) {
+          return res.send(">>> Error: " + err.toString() + ".");
+        }
+
+        // Check for HTTP Errors
+        if (resp.statusCode !== 200) {
+          return res.send("> HTTP Error ( " + resp.statusCode + ").");
+        }
+
+        var udResp = JSON.parse(body);
```
mitchmcdee: maybe give a description for this variable or choose a better name as its a bit unclear

scripts/urban.js
```diff
@@ -6,68 +6,49 @@
...
+            }
+            res.send(response);
+            if (udResp["list"].length > 1) {
+              res.send(" - more definitions at http://www.urbandictionary.com/define.php?term=" + encodeURI(phrase));
+            }
+            return;
+          } catch(err) {
+            return res.send(">>> Error parsing Urban Dictionary response: " + err.toString());
+          }
+        } else {
```
mitchmcdee: check the negative condition here and remove a level of indentation :^) will read nicer

i.e. `if (udResp["result_type"] == undefined || udResp["result_type"] == "no_results")` or w/e (is there a successful result type? maybe just check for that)

scripts/urban.js
```diff
@@ -6,68 +6,49 @@
...
+        // Check for HTTP Errors
+        if (resp.statusCode !== 200) {
+          return res.send("> HTTP Error ( " + resp.statusCode + ").");
+        }
+
+        var udResp = JSON.parse(body);
+
+        // Parse Urban Dictionary response and send result.
+        if (udResp["result_type"] !== undefined && udResp["result_type"] !== "no_results") {
+          try {
```
mitchmcdee: why try catch here? seems lazy hehe ;)

scripts/urban.js
```diff
@@ -6,68 +6,49 @@
...
+    var phrase = res.match[1].trim();
+    var urban = robot.http("http://api.urbandictionary.com/v0/define?term=" + encodeURI(phrase))
+      .get()(function (err, resp, body) {
+
+        // Check for correct usage
+        if (res.match[1][0] !== " " || !phrase || phrase.length < 1) {
+          return res.send("> Usage: `!urban <SEARCH_PHRASE>`");
+        }
+
+        // Check for JS/hubot errors
```
mitchmcdee: fix this description, I think it actually returns an error if the `get()` errors out i.e. no response? check the documentation to be sure

scripts/urban.js
```diff
@@ -6,68 +6,49 @@
...
-		json = JSON.parse(string);
-		return json["list"][0]["definition"];
-	};
-
-	// Parses the JSON to get the first definition
-	robot.getUrbanExample = function (string) {
-		json = JSON.parse(string);
-		return json["list"][0]["example"];
-	};
+  robot.respond(/!?urban(.*)/i, function (res) {
```
mitchmcdee: I know you trim anyway, but maybe chuck a space after urban just for clarity sake? (unless there's a reason?)

scripts/urban.js
mitchmcdee: (just make it more clear why this error would occur, as "JS" isn't accurate)

scripts/urban.js
nickl93: This actually worked out really nicely as the UD api will always return those properties if a result is found. Thanks :).

scripts/urban.js
```diff
@@ -0,0 +1,50 @@
...
+//
+
+module.exports = function (robot) {
+  robot.respond(/!?urban(.*)/i, function (res) {
+    var phrase = res.match[1].trim();
+    var urban = robot.http("http://api.urbandictionary.com/v0/define?term=" + encodeURI(phrase))
+      .get()(function (err, resp, body) {
+
+        // Check for correct usage
+        if (res.match[1][0] !== " " || !phrase || phrase.length < 1) {
```
mitchmcdee: you won't need to check for whitespace if you fix the regex. Also, this if statement should be moved to before the `get()` request so we can validate it :)

mitchmcdee: you won't need to check for whitespace if you fix the regex. Also, this if statement should be moved to before the get() request so we can validate it :)

scripts/urban.js
```diff
@@ -28,23 +30,23 @@ module.exports = function (robot) {
...
+        // Parse Urban Dictionary response and send result.
+        var firstResult = udResp["list"][0];
+        var definition = firstResult["definition"];
+        var example = firstResult["example"];
+        var response = phrase.toUpperCase() + ":\n" + definition.toString() + "\n";
+        if (example) {
+          response += ">>> " + example.toString();
+        }
+        res.send(response);
+        if (udResp["list"].length > 1) {
```
mitchmcdee: I would put this a bit higher and `+=` it onto the response so you can send it as one message :^)

scripts/urban.js
```diff
@@ -6,16 +6,18 @@
 //
 
 module.exports = function (robot) {
-  robot.respond(/!?urban(.*)/i, function (res) {
+  robot.respond(/!?urban( .+)?/i, function (res) {
```
mitchmcdee: I think this should this be `/!?urban (.+)/i`, removes the whitespace and makes input non-optional :)

gricey432: But then it wouldn't match and go through to the help text if someone uses it without args

mitchmcdee: ah yep true :)

nickl93: Yeah, I spent a good hour learning regex to get it to do the thing properly. :) Figured we'd want it to respond to no input. If I have time/remember (and you guys approve) I'll go back over the other scripts and get them to do the same thing.

scripts/urban.js
```diff
@@ -0,0 +1,52 @@
...
+        // Parse Urban Dictionary response and send result.
+        var firstResult = udResp["list"][0];
+        var definition = firstResult["definition"];
+        var example = firstResult["example"];
+        var response = phrase.toUpperCase() + ":\n" + definition.toString() + "\n";
+        if (example) {
+          response += ">>> " + example.toString();
+        }
+        res.send(response);
+        if (udResp["list"].length > 1) {
```
mitchmcdee: I'm not sure if this was in my last review/if github goofed, but you can move this if check to above the `res.send` and `+=` it to response to send it all as one message :)

nickl93: @mitchmcdee is it better for me to split parse every line of the example adding ‘>’  to the start (like Matthew's original one), or sending two messages? Or is there another fix I'm not seeing?

mitchmcdee: image
this is another reason we should avoid multiple sends, race conditions :^)

lemme poke around with your code and see what we can do

mitchmcdee:
```
var response = phrase.toUpperCase() + ":\n" + definition + "\n";
if (example) {
  response += ">>>" + example;
}
var more = "";
if (udResp["list"].length > 1) {
  more = "_more definitions at http://www.urbandictionary.com/define.php?term=" + encodeURI(phrase) + ' _';
}
res.send(response, more);
```

I think this should satisfy things
was there any reason you had the toString()'s there? I don't think they were necessary since the response is just a string?

nickl93: So clean, thank you! Yes, toStrings were unnecessary, have removed. :)

PR #212, opened 2017-11-12
Ensure that any new default stats being added are added to the brain

mitchmcdee:
Subscription system will currently be broken as `_subscriber` was never added to the brain, and no code was in place to check otherwise.

This PR fixes that (and any future stats added to the system) by checking all default stats are present every time we get stats from the database.

Also fixes a mutation bug (`DEFAULT_STATS` was copied by reference and therefore being mutated-- not a big deal for now but probably not a good thing lol)

Note: this bug slipped through the gaps during testing because I had no database locally-- and this is only an issue to do with persistent storage

### DIFF
--- a/scripts/stats.js
+++ b/scripts/stats.js
@@ -8,7 +8,7 @@
 var HubotCron = require('hubot-cronjob');
 
 // Default stats upon reset
-DEFAULT_STATS = {rooms: {}, commands: {}, _subscribers: {}};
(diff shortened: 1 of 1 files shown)


mitchmcdee:
Changed the way I did it @TRManderson to be more simple + preserve old stats instead of wiping them if it was discovered a default stat was not present, so PTAL again when you have the time :^)

PR #216, opened 2017-11-14
Prevent wavie.js executing for #general (which has its own wave)

mitchmcdee:
Fixes the "double wave" situation


CameronAavik:
@TRManderson reckon we merge this?

CameronAavik:
Actually, I'll do it anyways 😂

PR #217, opened 2017-11-15
Created !attic for ez retrieval of UQAttic course files

mitchmcdee:
- Will use current channel as the course code if none is provided
- Retrieves all files from the course folder and sends it back to room
- If there are >15 files, will instead send it to the user to prevent flooding the room

As always, feedback and h@rdc0re review is much appreciated :^)


mitchmcdee:
Course room:
image

Regular usage:
image
image
image

mitchmcdee:
Closing as this is a) unreviewed and b) is for the jsbot. Instead, I will create a new PR with the ported python code

PR #221, opened 2018-01-20
First version of rewrite in python

CameronAavik:
Am merging this into a branch called python-rewrite for now until it has feature parity.

Some notable things

- Added MIT License
- No bot frameworks any more
- I have already implemented the commands cat, dog, spider, radar, repo, and cookbook
- README has been updated with all the new setup instructions. It is pretty different from the last one so would appreciate it if someone can test out going through the steps and seeing if they can get the bot to run.
- I'm going to try and make it such that the API wrapper will return objects instead of the ids. So for example, when receiving a command the API returns a Channel object which you could call `.get_members()` on as opposed to just being given a channel id. 
- Ideally, the scripts folder should never need to see the `SlackClient` object.
- Everything should have type hints (if you see things without them, let me know).
- Pagination is already implemented, check out `api.py` and `Channel.get_members` to see. The docs claim that every single API call which is paginated will follow the same structure and should be supported by what I've written.
- Tom wrote a thing which makes it so that every python file in the `uqcsbot/scripts` directory will get loaded. This makes it possible to add new commands by only adding one file. 
- I copied all the commands from the current uqcsbot into the `unimplemented_commands` directory so we can remove it as we implement things, it will also make it easy in the code reviews/pull requests to compare the code and check that it does the same thing.

All comments/criticisms are welcome, also it should be at the point now where others should be able to start contributing to it and porting over commands.


TRManderson:
By using normal `flask.start` in `uqcsbot.main` we're going to always have a shit webserver that errors constantly, as it'll use the reference WSGI implementation. At a bare minimum, we should add `waitress` as a dep and use `waitress.serve`.

Also I'm not really OK with instantiating the events API object and the slack client object in `uqcsbot.base`. I feel as though they should be instantiated in our `main` function, passed into the command handler in some `run` function, and then passed into handler functions as arguments (or as context local variables ala flask globals)

CameronAavik:
@TRManderson 
I had no clue that the default WSGI implemenation was buggy, I'd be fine switching to waitress as long as people writing commands don't need to worry about it being there

Also yes you are right, it would be better to have those objects in main rather than in base. It also allow for easy mocking/stubbing

PR #222, opened 2018-01-28
Add asyncio support

TRManderson:
Runs all event handlers asynchronously. If a worker is written synchronously, runs it in a thread pool executor, else uses standard asyncio.

Using mobile data and github is being weird so this might be a bad merge.

Currently untested.

### DIFF
--- a/uqcsbot/base.py
+++ b/uqcsbot/base.py
@@ -4,8 +4,11 @@ from slackeventsapi.server import SlackServer
 from .api import Channel
 import waitress
 import collections
+import asyncio
+import concurrent.futures
 from typing import Callable
 
+CommandHandler = Callable[[Command], None]
 
 class Command(object):
     def __init__(self, command_name: str, arg: str, channel: Channel):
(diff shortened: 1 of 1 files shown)


uqcsbot/base.py
```diff
@@ -39,10 +50,12 @@ def _handle_command(self, event_data: dict) -> None:
         channel = Channel(self.client, message["channel"])
         command = Command(command_name, None if not arg else arg[0], channel)
         for cmd in self._command_registry[command_name]:
-            cmd(command)
+            asyncio.ensure_future(cmd(command), self.evt_loop)
 
     def on_command(self, command_name: str):
         def decorator(fn):
+            if not asyncio.iscoroutinefunction(fn):
```
CameronAavik: Just to confirm that I understand this correctly, is this code here to allow you to optionally make async scripts?

TRManderson: It allows you to _write_ it optionally async, everything will be run async though

PR #225, opened 2018-02-04
API wrapper - Allow calling API through standard dotted methods

TRManderson:
Better documentation than most of my code, but still untested as I have very limited internet access.

Simplifies API calls a bit, and makes the API look cool :joy:

### DIFF
--- a/uqcsbot/api.py
+++ b/uqcsbot/api.py
@@ -4,19 +4,15 @@ from slackclient import SlackClient
 import asyncio
 
 
-def paginate_api_call(client: SlackClient, method: str, **kwargs) -> List[dict]:
-    return list(Paginator(client, method, **kwargs))
-
-
 class Channel(object):
     def __init__(self, client: SlackClient, channel_id: str):
         self.client = client
+        self.api = APIWrapper(self.client)
         self.id = channel_id
 
     def get_members(self) -> List[str]:
-        member_pages = paginate_api_call(self.client, "conversations.members", channel=self.id)
         members = []
-        for page in member_pages:
+        for page in self.api.conversations.members.paginate(channel=self.id):
             members += page["members"]
         return members
 
--- a/uqcsbot/api.py
+++ b/uqcsbot/api.py
@@ -1,17 +1,12 @@
-from typing import List, Union, Awaitable
+from typing import List, Iterable, AsyncIterable, AsyncGenerator, Generator, Any
 from functools import partial
 from slackclient import SlackClient
 import asyncio
 
 
 # see https://api.slack.com/docs/pagination for details
-def paginate_api_call(client: SlackClient, method: str, *args, **kwargs) -> List[dict]:
-    pages = []
-    while kwargs.get("cursor") != "":
-        page = client.api_call(method, *args, **kwargs)
-        pages.append(page)
-        kwargs["cursor"] = page["response_metadata"]["next_cursor"]
-    return pages
+def paginate_api_call(client: SlackClient, method: str, **kwargs) -> List[dict]:
+    return list(Paginator(client, method, **kwargs))
 
 
 class Channel(object):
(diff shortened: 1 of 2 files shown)


uqcsbot/api.py
```diff
@@ -24,3 +26,83 @@ def get_members(self) -> List[str]:
...
+            client=self._client,
+            method=item,
+            is_async=self._async
+        )
+
+    def __call__(self, **kwargs) -> 'APIWrapper':
+        """
+        On call, reconfigure the API Wrapper
+        """
+        return type(self)(self._client, **kwargs)
```
CameronAavik: What is this doing?

uqcsbot/api.py
```diff
@@ -1,5 +1,7 @@
-from typing import List
+from typing import List, Union, Awaitable
+from functools import partial
 from slackclient import SlackClient
+import asyncio
 
 
 # see https://api.slack.com/docs/pagination for details
```
CameronAavik: For some reason github can't let me comment on it, but can you update the paginate_api_call while you are here?

CameronAavik: Or at least somehow move pagination into the API Wrapper.

uqcsbot/api.py
```diff
@@ -24,3 +26,83 @@ def get_members(self) -> List[str]:
...
+class APIMethodProxy(object):
+    """
+    Helper class used to implement APIWrapper
+    """
+    def __init__(self, client: SlackClient, method: str, is_async: bool = False):
+        self._client = client
+        self._method = method
+        self._async = is_async
+
+    def __call__(self, timeout=None, **kwargs) -> Union[dict, Awaitable[dict]]:
```
CameronAavik: Can you add documentation for the timeout parameter? What happens when it times out?

uqcsbot/api.py
TRManderson: This is what enables the following example:
```python
api = APIWrapper(client)
async_api = api(is_async=True)
await async_api.chat.postMessage(channel="general", text="message")
```

It constructs a new object of the same type, passing in the same SlackClient, but otherwise requiring you to pass in any other config. The current behaviour is slightly different to what I intended so I'll update both the code & docs.

uqcsbot/api.py
TRManderson: The timeout parameter is just to replicate the ability of `SlackClient.api_call` to pass in timeout as a positional argument. I could remove `timeout=None,` and this would function effectively the same. Just gets passed straight through to `SlackClient.api_call`

Happy to still document it if you want

CameronAavik: Ah ok, if you can that'd be good but I'm not too fussed about it if it's just that.

TRManderson: Just changed it to `**kwargs` instead

uqcsbot/api.py
TRManderson: Made a `Paginator` class that does iteration (async or otherwise) which is cool

uqcsbot/api.py
```diff
@@ -1,15 +1,12 @@
...
 
 # see https://api.slack.com/docs/pagination for details
-def paginate_api_call(client: SlackClient, method: str, *args, **kwargs) -> List[dict]:
-    pages = []
-    while kwargs.get("cursor") != "":
-        page = client.api_call(method, *args, **kwargs)
-        pages.append(page)
-        kwargs["cursor"] = page["response_metadata"]["next_cursor"]
-    return pages
+def paginate_api_call(client: SlackClient, method: str, **kwargs) -> List[dict]:
```
CameronAavik: I don't think we need this method any more, I believe we can replace the usage in `get_members` with `api.conversations.members.paginate(channel=self.id)`

This change doesn't block merging in though, in case you wanted to leave it for later.

PR #227, opened 2018-02-04
First pass at using RTM client

TRManderson:
Again this is untested code. Hopefully works fine. It at least removes a few deps.

### DIFF
--- a/uqcsbot/__init__.py
+++ b/uqcsbot/__init__.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import importlib
+import logging
 from uqcsbot.base import UQCSBot, bot, Command
 
 SLACK_VERIFICATION_TOKEN = os.environ.get("SLACK_VERIFICATION_TOKEN", "")
@@ -19,9 +20,11 @@ def main():
 
     # Run bot
     if '--dev' in sys.argv:
+        logging.basicConfig(level=logging.DEBUG)
         bot.run_debug()
     else:
-        bot.run(SLACK_BOT_TOKEN, SLACK_VERIFICATION_TOKEN, host='0.0.0.0', port=5000)
+        logging.basicConfig(level=logging.INFO)
+        bot.run(SLACK_BOT_TOKEN, SLACK_VERIFICATION_TOKEN)
 
 if __name__ == "__main__":
     main()
--- a/uqcsbot/__init__.py
+++ b/uqcsbot/__init__.py
@@ -19,6 +19,7 @@ def main():
         importlib.import_module(module)
 
     # Run bot
+    # TODO: Make logging command-line configurable
     if '--dev' in sys.argv:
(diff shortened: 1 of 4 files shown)


uqcsbot/__init__.py
```diff
@@ -19,9 +20,11 @@ def main():
 
     # Run bot
     if '--dev' in sys.argv:
+        logging.basicConfig(level=logging.DEBUG)
```
CameronAavik: I would rather the logging level toggle be a separate command, I do all the debugging directly against the slack API and would like the debug level in that circumstance

uqcsbot/base.py
```diff
@@ -74,12 +83,25 @@ def _handle_command(self, event_data: dict) -> None:
 
     def on_command(self, command_name: str):
         def decorator(fn):
-            if not asyncio.iscoroutinefunction(fn):
```
CameronAavik: Is on_command being deprecated? I see this code got moved to `register_handler` instead.

uqcsbot/base.py
```diff
@@ -41,28 +39,39 @@ def async_worker(loop):
...
 
-    def _handle_command(self, event_data: dict) -> None:
-        message = event_data["event"]
+        self._handlers[''].append(self._handle_command)
+        self._handlers['hello'].append(self._handle_hello)
+        self._handlers['hello'].append(self._handle_goodbye)
+
+    def _handle_hello(self, evt):
+        if evt != {"type": "hello"}:
+            self.logger.debug(f"Goodbye event has unexpected extras: {evt}")
```
CameronAavik: log should say "Hello" instead of "Goodbye"

uqcsbot/base.py
TRManderson: `on_command` is specific to the `Command` wrapper, and I think having such a wrapper is a really good idea. `on` (which uses `register_handler`), however, can be used to subscribe to any events whatsoever. So no, not deprecated, this is just a more specific API

uqcsbot/__init__.py
TRManderson: You OK with this being provisional and moving it out once we do nicer arg parsing? (perhaps with `click`). If so, I'll add a `# TODO`, if not I'll change it now.

CameronAavik: Am fine with fixing it later.

uqcsbot/base.py
CameronAavik: How do non-async functions get called now? Since you are no longer wrapping them with run_async.

TRManderson: Rip, was intending to replace this differently but evidently I didn't. Code review catches bugs yet again!

TRManderson: Fixed

TRManderson:
Actually tested now!

PR #229, opened 2018-02-12
Wavie.js

TRManderson:
Implement wavie.js in Python


CameronAavik:
I was kinda hoping to write the commands in such a way that they don't directly need to know how slack's API works. thus making it easy for others to write commands that use the same methods. Would we be able to abstract out the `bot.on('message')`, `channels.find` and `reactions.add` into functions that can be reused elsewhere?

TRManderson:
I think we can change `bot.client.server.channels` to `bot.channels` fairly trivially. I'm honestly not too opposed to `bot.api.reactions.add`

jack775544:
I partially agree Cameron on the idea that there should be some helpers to abstract things away, but I would say that you would have to be careful to not get an ungoogleable mess at the end of it.

CameronAavik:
actually you're right, `reactions.add` isn't too bad. I'm mostly wanting the `bot.on('message')` to change since from reading the script it is unclear why I'd be listening to a message if I want to know someone joined.

TRManderson:
I think adding more `on_blah` methods is a very good idea and we definitely should do that. Not sure what the best approach here is though - there's a `channel_join` event and an `channel_join` message and I want one but not the other - how do we distinguish between the two with the handler names?

jack775544:
Well it isn't lying, it is listening to a message, but that message is a slack message that indicates weather a person joins or leaves. Personally I think that is very reasonable. IMO the most complex part there which should be abstracted out is the message channel find that was done.

TRManderson:
Yeah, I've got a PR with the channel find stuff

CameronAavik:
When would you want to use the event vs using the message?

jack775544:
You would use event in the case of something like welcome. Listening to the message is done here since it wants to react to the message that is placed in slack.

CameronAavik:
how about `on_channel_join_message` and `on_channel_join_event`?

jack775544:
I would say leave the decorator as is and just have an enum defined of all the events you can listen to. That way you could do something like `@bot.on(Uqcsbot.Events.MESSAGE)` or something along those lines

TRManderson:
See #230

TRManderson:
Updated to use new channel cache wrapper - merge that first

PR #230, opened 2018-02-12
Use the Slack-internal channel list

TRManderson:
Uses `slackclient`'s `Channel` representation

### DIFF
--- a/uqcsbot/api.py
+++ b/uqcsbot/api.py
@@ -4,19 +4,6 @@ from slackclient import SlackClient
 import asyncio
 
 
-class Channel(object):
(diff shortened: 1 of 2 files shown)


uqcsbot/api.py
```diff
@@ -4,19 +4,6 @@
 import asyncio
 
 
-class Channel(object):
-    def __init__(self, client: SlackClient, channel_id: str):
-        self.client = client
-        self.api = APIWrapper(self.client)
-        self.id = channel_id
-
-    def get_members(self) -> List[str]:
```
TRManderson: replaced by channel.members

PR #233, opened 2018-02-12
Added cronjob event handling support

CameronAavik:
Doesn't actually do wakie implementation while tom writes channel wrapper.
It even has async support.


setup.py
```diff
@@ -10,7 +10,8 @@
 install_requires = [
     'slackclient',
     'requests',
-    'BeautifulSoup4'
+    'BeautifulSoup4',
+    'apscheduler'
```
TRManderson: Just add trailing commas anyway

PR #234, opened 2018-02-12
Channel cache/wrapper

TRManderson:


uqcsbot/api.py
```diff
@@ -133,4 +138,137 @@ def __call__(self, **kwargs) -> 'APIWrapper':
...
+    def get(self, name_or_id: str, default: T = None, use_cache: bool =True) -> Union[Channel, T]:
+        if use_cache and not self._initialised and:
+            self._initialise()
+        if name_or_id in self._channels_by_id:
+            return self._channels_by_id[name_or_id]
+        elif name_or_id in self._channels_by_name:
+            return self._channels_by_name[name_or_id]
+        elif not use_cache:
+            with self._lock:
+                resp = self._bot.api.channels.info(channel=name_or_id)
```
CameronAavik: A quick check of the API says that this endpoint takes ID, not name https://api.slack.com/methods/channels.info

uqcsbot/api.py
```diff
@@ -133,4 +138,137 @@ def __call__(self, **kwargs) -> 'APIWrapper':
...
+        return chan
+
+    def _initialise(self):
+        with self._lock:
+            for page in self._bot.api.channels.list.paginate():
+                for chan in page['channels']:
+                    self._add_channel(chan)
+            self._initialised = True
+
+    def get(self, name_or_id: str, default: T = None, use_cache: bool =True) -> Union[Channel, T]:
```
CameronAavik: I'd rather have two named parameters, name and id than a joint variable here I think. This also means that we can throw do something different depending on if they passed in name or id when they said use_cache = False

uqcsbot/api.py
TRManderson: Yep. This shoud be ID in this case. There's no easy way for me to check.

uqcsbot/api.py
TRManderson: I'm trying to replicate `dict` behaviour here with `.get`

CameronAavik: Hmm, fair enough. Maybe add another method then called `invalidate_caches` so that the user has someway of forcing it to reload data for channels by name.

TRManderson:
Now has been tested and should be good to go, updating #229 based on this

PR #242, opened 2018-02-18
Add welcome

TRManderson:


uqcsbot/base.py
```diff
@@ -217,6 +219,7 @@ def run(self, api_token, verification_token, **kwargs):
                     self._run_handlers(message)
                     if message.get('type') == "goodbye":
                         break
+                time.sleep(0.5)
```
CameronAavik: what's the sleep for?

uqcsbot/scripts/welcome.py
```diff
@@ -0,0 +1,39 @@
+# Welcomes new users to UQCS Slack and check for member milestones
+from uqcsbot import bot
+import asyncio
+
+MEMBER_MILESTONE = 50;  # Number of members between posting a celebration
+MESSAGE_PAUSE = 2500 / 1000;   # Number of seconds between sending bot messages
```
CameronAavik: Can we just put 2.5 here? It took be a bit to realise that you were doing milliseconds then converting to seconds.

uqcsbot/scripts/welcome.py
```diff
@@ -0,0 +1,39 @@
...
+];
+
+@bot.on("member_joined_channel")
+async def welcome(evt: dict):
+    chan = bot.channels.get(evt.get('channel'))
+    if chan is None or chan.name != "announcements":
+        return
+    announcements = chan
+    general = bot.channels.get("general")
+    user_info = await bot.run_async(bot.api.users.info, user=evt.get("user"))
```
CameronAavik: might be good to use the users cache instead of an API call here. I know it's not done yet though, so we can do that later

uqcsbot/scripts/welcome.py
```diff
@@ -0,0 +1,39 @@
...
+    display_name = user_info.get("user", {}).get("profile", {}).get("display_name")
```
CameronAavik: If the display_name ends up being null here, we should not try and post a welcome message I think saying `"Welcome, None"`

uqcsbot/scripts/welcome.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # Ensure the message to #general is scheduled, but don't block
```
CameronAavik: I don't think there is any harm in doing the two synchronously here, seeing all this async stuff in the code is already making it somewhat beginner unfriendly.

uqcsbot/base.py
TRManderson: We were busy waiting

uqcsbot/scripts/welcome.py
TRManderson: Yeah fair, I just copy-pasted from the old code where JS uses milliseconds

uqcsbot/scripts/welcome.py
TRManderson: Yeah, I'm debating whether or not to implement it right now. Ideally we'd be using the users cache.

uqcsbot/scripts/welcome.py
TRManderson: Yeah I'm undeniably going overboard with the async. It all runs in a thread anyway so I should probs tone it down a bit 😂

PR #243, opened 2018-02-19
!ecp and !events command

CameronAavik:


CameronAavik:
I've also since learnt not to rebase onto master but instead to merge onto master. Thus why you see the merge remote commits in this PR :'(

CameronAavik:
Woops, I thought this was already merged in. I've put !events on top of this as well, could it be reviewed again @TRManderson

mitchmcdee:
maybe split it out?

CameronAavik:
I am not that good at git, I should be making branches for each command, I'll do that in the future

mitchmcdee:
fair fair hehe

uqcsbot/scripts/events.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class EventFilter(object):
+    def __init__(self, full=False, weeks=None, cap=None, is_valid=True):
+        self.is_valid = is_valid
+        self._full = full
+        self._weeks = weeks
+        self._cap = cap
+
+    @staticmethod
```
TRManderson: It doesn't really matter in this case but pls use `@classmethod`

uqcsbot/scripts/events.py
```diff
@@ -0,0 +1,120 @@
...
+    def from_command(command: Command):
+        if not command.has_arg():
+            return EventFilter(weeks=2)
```
TRManderson: Use classmethod and construct with `cls`

uqcsbot/scripts/events.py
```diff
@@ -0,0 +1,120 @@
+from typing import List
+from uqcsbot import bot, Command
+from icalendar import Calendar, vText
+from requests import get
```
TRManderson: importing `requests` plain and using `requests.get` is more common - `get` as a plain verb is ambiguous when reading later code.

It might be good to also look into us doing HTTP sessions properly, not in the scope of this PR though

uqcsbot/scripts/events.py
```diff
@@ -0,0 +1,120 @@
...
+from datetime import datetime, timedelta
+from pytz import timezone, utc
```
TRManderson: Please add `pytz` to our explicit dependencies

uqcsbot/scripts/events.py
```diff
@@ -0,0 +1,120 @@
...
+    http_response = await bot.run_async(get, CALENDAR_URL)
+    cal = Calendar.from_ical(http_response.content)
+
+    current_time = datetime.now(tz=BRISBANE_TZ).astimezone(utc)
+
+    # TODO: support recurring events
+    # subcomponents are how icalendar returns the list of things in the calendar
+    # we are only interested in ones with the name VEVENT as they are events
+    # we also currently filter out recurring events
+    events = [Event.from_cal_event(c) for c in cal.subcomponents if c.name == 'VEVENT' and c.get('RRULE') is None]
```
TRManderson: I don't really like the list comprehensions here given they're creating all these intermediate lists, but also it's not enough of a perf worry to bother changing it

uqcsbot/scripts/events.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class Event(object):
+    def __init__(self, start: datetime, end: datetime, location: vText, summary: vText):
+        self.start = start
+        self.end = end
+        self.location = location
+        self.summary = summary
+
+    @staticmethod
```
TRManderson: classmethod pls

joeladdison:
If you want to split it out:
`$ git branch convert-events 904e762`
`$ git reset HEAD~2`
`$ git push --force`
then push up the convert-events later branch when ready.
I'm sure there are other ways, but that sequence should give you two branches, and two PRs later.

TRManderson:
woah @CameronAavik I didn't mean to unroll the list comp into a for loop, I meant given you were creating several intermediate lists you could just use a generator expression

CameronAavik:
The last two list transformations act on the entire list, not per item so I couldn't do a list comp. I only combined the two list comprehensions into a for loop as I thought that the first one was getting a bit long.

Also I think it's more readable personally

TRManderson:
Yeah it is more readable, and I'm fine with it, hence the approval

PR #246, opened 2018-02-26
Make js wakie less retarded

nick-potts:


scripts/wakie.js
```diff
@@ -16,9 +16,9 @@ module.exports = function(robot) {
 
         // Send ping to two random victims in #general and react with random emoji
         var general = robot.adapter.client.rtm.dataStore.getChannelByName("general");
-        var members = general.members.filter(user => !user.deleted);
-        var victim1 = members[Math.floor(Math.random() * members.length)];
-        var victim2 = members[Math.floor(Math.random() * members.length)];
+        var members = general.members.filter(user => !user.deleted && !user.is_bot);
```
CameronAavik: I'm actually concerned that the reason it returns uqcsbot twice so frequently is because this list only contains 1 item, so filtering this out would leave an empty list of members.

scripts/wakie.js
```diff
@@ -16,9 +16,9 @@ module.exports = function(robot) {
...
+        var victim1 = members[Math.floor(Math.random() * members.length * 0.5)];
+        var victim2 = members[Math.floor(Math.random() * members.length * 0.5 + (members.length / 2))];
```
mitchmcdee: this is just `Math.floor(Math.random() * members.length)`

scripts/wakie.js
```diff
@@ -16,9 +16,9 @@ module.exports = function(robot) {
 
         // Send ping to two random victims in #general and react with random emoji
         var general = robot.adapter.client.rtm.dataStore.getChannelByName("general");
-        var members = general.members.filter(user => !user.deleted);
-        var victim1 = members[Math.floor(Math.random() * members.length)];
-        var victim2 = members[Math.floor(Math.random() * members.length)];
+        var members = general.members.filter(user => !user.deleted && !user.is_bot);
+        var victim1 = members[Math.floor(Math.random() * members.length * 0.5)];
```
mitchmcdee: What's the rationale behind this?

scripts/wakie.js
CameronAavik: I think the code was meant to be

    Math.floor(Math.random() * members.length * 0.5) + (members.length / 2)

So that the first member would be from the first half of the list and the second member would be from the second half

CameronAavik:
Closing as it does not fix anything.

PR #253, opened 2018-03-04
Added !wiki command

Prometheus9103:
Works just like the js version of !wiki though it only triggers on !wiki and not !wikipedia etc.

Could also add the link to the result and format it in a code block?


uqcsbot/scripts/wiki.py
```diff
@@ -0,0 +1,26 @@
+from uqcsbot import bot, Command
+import requests
+import json
+
+
+@bot.on_command("wiki")
+async def handle_wiki(command: Command):
+    search = command.arg
```
mitchmcdee: can probably have a better variable name than 'search', maybe 'search_query'?

uqcsbot/scripts/wiki.py
```diff
@@ -0,0 +1,26 @@
...
+    url = f"https://en.wikipedia.org/w/api.php?action=opensearch&search={search}&format=json"
+    http_response = await bot.run_async(requests.get, url)
+    response = json.loads(http_response.content)
+
+    # If the results are empty let them know
+    try:
+        info = response[2][0]
```
draganmarjanovic: I'd personally chuck the indexes into descriptive vars if possible.

uqcsbot/scripts/wiki.py
```diff
@@ -0,0 +1,26 @@
+from uqcsbot import bot, Command
+import requests
+import json
+
+
+@bot.on_command("wiki")
+async def handle_wiki(command: Command):
+    search = command.arg
+    url = f"https://en.wikipedia.org/w/api.php?action=opensearch&search={search}&format=json"
```
mitchmcdee: using fstrings here seems awkward-- I think it'd be better to pass 'search' as a kwarg like here: http://docs.python-requests.org/en/master/user/quickstart/#passing-parameters-in-urls, just inline is fine

uqcsbot/scripts/wiki.py
mitchmcdee: what's the 2 and 0 here mean? give some description :)

also, it'd be better to do this in an if statement to remove an unnecessary try/except

uqcsbot/scripts/wiki.py
```diff
@@ -0,0 +1,26 @@
...
+    http_response = await bot.run_async(requests.get, url)
+    response = json.loads(http_response.content)
+
+    # If the results are empty let them know
+    try:
+        info = response[2][0]
+    except IndexError:
+        bot.post_message(command.channel, "No Results.\r\n")
```
mitchmcdee: I don't think we need the `\r\n` if we're just posting the message

uqcsbot/scripts/wiki.py
```diff
@@ -0,0 +1,26 @@
...
+        return
+
+    # If it gives suggestions just return the first one
+    typical = "may refer to:"
+    if typical in info:
```
mitchmcdee: we should return something here if typical is not in info

mitchmcdee:
Oh, and be sure to include a screenshot or something of the various outputs :)

mitchmcdee:
It looks like this can be simplified and remove a lot of confusion/complexity if we use something like limit=1 to do something like:

```
search_query = command.arg
api_url = 'https://en.wikipedia.org/w/api.php?action=opensearch&limit=1'
http_response = await bot.run_async(requests.get, api_url, params={'search': search_query})
_, title_list, snippet_list, url_list = json.loads(http_response.content)

# All lists will be of the same length, so we can just check one of them to see if we have any results
if len(title_list) == 0:
    bot.post_message(command.channel, "No results found.")
    return

# Get the first result's title, snippet text and url
title, snippet, url = title_list[0], snippet_list[0], url_list[0]
bot.post_message(command.channel, f'{title}: {snippet}\nLink: {url}')
```

also, what's the 'suggestion' stuff do? I left that out because I wasn't sure
also this code is completely untested but hopefully conveys my point
also you should check that the http_response is valid before doing any json parsing I believe

Prometheus9103:
That's a good point. What I had was basically just a copy of the original wiki command. 

The suggestion stuff is because sometimes the first result isn't actually any content it is just some variation of the phrase "it may refer to" and the following results are the potential matches.

mitchmcdee:
totally understand :^)

and I see, could you post an example response? just curiouz :)

Prometheus9103:
https://en.wikipedia.org/w/api.php?action=opensearch&format=json&search=wakanda
and
https://en.wikipedia.org/w/api.php?action=opensearch&format=json&search=those

are two variations that I have found

mitchmcdee:
lit. Okay so maybe instead you can do:

```
search_query = command.arg
# Limit to 2 results to account queries that could refer to multiple things (e.g. Wakanda)
api_url = 'https://en.wikipedia.org/w/api.php?action=opensearch&limit=2'
http_response = await bot.run_async(requests.get, api_url, params={'search': search_query})
_, title_list, snippet_list, url_list = json.loads(http_response.content)

# All lists will be of the same length, so we can just check one of them to see if we have any results
if len(title_list) == 0:
    bot.post_message(command.channel, "No results found.")
    return

# Get the first result's title, snippet text and url
title, snippet, url = title_list[0], snippet_list[0], url_list[0]
# Detect if there are multiple references to the query, if so, use the first reference (i.e. the second item in the lists).
multiple_reference_instances = ("may refer to:", "may have several meanings:")
if any(instance in snippet for instance in multiple_reference_instances):
   title, snippet, url = title_list[1], snippet_list[1], url_list[1]

bot.post_message(command.channel, f'{title}: {snippet}\nLink: {url}')
```

Prometheus9103:
That works.

Should I just check if the requests status code is ok before proceeding or do more?

mitchmcdee:
sounds sufficient to me :^) but maybe double check the status codes of the API (it still might return OK even if it ain't-- I'm not sure)

uqcsbot/scripts/wiki.py
```diff
@@ -0,0 +1,30 @@
...
+
+    # If the results are empty let them know. Any list being empty signifies this.
+    if len(title_list) == 0:
+        bot.post_message(command.channel, "No Results Found.")
+        return
+
+    title, snippet, url = title_list[0], snippet_list[0], url_list[0]
+    # Detect if there are multiple references to the query, if so, use the first reference (i.e. the second item in the lists).
+    multiple_reference_instances = ("may refer to:", "may have several meanings:")
+    if any(instance in snippet for instance in multiple_reference_instances) or not snippet:
```
mitchmcdee: taking the second reference if snippet is empty doesn't really make sense, as this if condition is entirely about finding if the response contains multiple reference.

Can you provide a link to an example that produces an empty snippet? we can decide what the best action is from there I guess

Prometheus9103:
Yeah I agree but I had a weird fluke example  with the query "mice and men" I think it was. Not sure why it happened.

mitchmcdee:
Ah yep I see. Let's just detect if any of the title/snippet/url are empty and return an error for now.

Prometheus9103:
Just for clarification do you mean to post  message stating the issue?

mitchmcdee:
correct :)

uqcsbot/scripts/wiki.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # If the results are empty let them know. Any list being empty signifies this.
+    if len(title_list) == 0:
+        bot.post_message(command.channel, "No Results Found.")
+        return
+
+    title, snippet, url = title_list[0], snippet_list[0], url_list[0]
+
+    # Sometimes the first element is an empty string which is weird so we handle that rare case here
+    if '' in (title, snippet, url):
```
mitchmcdee: this seems sketchy :^)
I would replace with something more verbose like `if len(title) == 0 or len(snippet) == 0 or len(url) == 0:`

Prometheus9103:
I also found another instance of the blank result problem: "mic". Not really sure what could be causing it

mitchmcdee:
I reckon it occurs when there is a page redirect for whatever reason.

mitchmcdee:
Good job on the script :^) lgtm

PR #257, opened 2018-03-06
Allows python bot to respond to commands in PMs

mubiquity:
Currently the channel list is only build using channels.list but this does not include the private messages the bot has with users. I just added the channels from the im.list. I also added a handler in the bot class which reloads all ims whenever a new team member joins (team_join event). This may cause an issue if the user immediately goes to the PM with the python bot and types a valid command. Not sure how to get around that.

Solution isn't particularly elegant and I am sure there is a lot to be improved.

### DIFF
--- a/uqcsbot/api.py
+++ b/uqcsbot/api.py
@@ -208,22 +208,6 @@ class ChannelWrapper(object):
         self._channels_by_id[chan.id] = chan
         return chan
 
-    def _add_im(self, im_dict: dict) -> Optional[Channel]:
-        if im_dict['is_user_deleted']:
-            return
-
-        # TODO: Name
-        chan = Channel(
-            bot=self._bot,
-            channel_id=im_dict['id'],
-            name='idk_what_you_want_here',
-            is_public=False,
-            is_private=True,
-            is_archived=False
-        )
-        self._channels_by_id[chan.id] = chan
-        return chan
-
     def _initialise(self):
         with self._lock:
             if self._initialised:
@@ -238,7 +222,12 @@ class ChannelWrapper(object):
 
             for page in self._bot.api.im.list.paginate():
                 for im in page['ims']:
-                    self._add_im(im)
+                    if im['is_user_deleted']:
+                        continue
+
+                    im['is_private'] = True
+                    im['name'] = im['id']
+                    self._add_channel(im)
 
             self._initialised = True
 
--- a/uqcsbot/api.py
+++ b/uqcsbot/api.py
@@ -208,6 +208,16 @@ class ChannelWrapper(object):
         self._channels_by_id[chan.id] = chan
         return chan
 
+    def reload_ims(self):
+        for page in self._bot.api.im.list.paginate():
+            for im in page['ims']:
+                if im['is_user_deleted']:
+                    continue
+
+                im['is_private'] = True
+                im['name'] = im['id']
+                self._add_channel(im)
+
     def _initialise(self):
         with self._lock:
             if self._initialised:
@@ -220,14 +230,7 @@ class ChannelWrapper(object):
                 for chan in page['channels']:
                     self._add_channel(chan)
 
-            for page in self._bot.api.im.list.paginate():
-                for im in page['ims']:
-                    if im['is_user_deleted']:
-                        continue
-
-                    im['is_private'] = True
-                    im['name'] = im['id']
-                    self._add_channel(im)
+            self.reload_ims()
 
             self._initialised = True
 
--- a/uqcsbot/api.py
+++ b/uqcsbot/api.py
@@ -2,7 +2,7 @@ from functools import partial
 from slackclient import SlackClient
 import asyncio
 import threading
-from typing import TYPE_CHECKING, List, Iterable, AsyncIterable, AsyncGenerator, Generator, Any, Union, TypeVar
+from typing import TYPE_CHECKING, List, Iterable, AsyncIterable, AsyncGenerator, Generator, Any, Union, TypeVar, Optional
 if TYPE_CHECKING:
     from .base import UQCSBot
 
@@ -208,6 +208,22 @@ class ChannelWrapper(object):
         self._channels_by_id[chan.id] = chan
         return chan
 
+    def _add_im(self, im_dict: dict) -> Optional[Channel]:
+        if im_dict['is_user_deleted']:
+            return
+
+        # TODO: Name
+        chan = Channel(
+            bot=self._bot,
+            channel_id=im_dict['id'],
+            name='idk_what_you_want_here',
+            is_public=False,
+            is_private=True,
+            is_archived=False
+        )
+        self._channels_by_id[chan.id] = chan
+        return chan
+
     def _initialise(self):
         with self._lock:
             if self._initialised:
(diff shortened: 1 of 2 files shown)


uqcsbot/base.py
```diff
@@ -81,6 +81,8 @@ def __init__(self, logger=None):
 
         self.channels = ChannelWrapper(self)
 
+        self.register_handler('team_join', self.channels.reload_ims)
```
mitchmcdee: This seems inefficient and inaccurate on first glance.
1) I don't know if IMs are created upon someone joining a team, I think they're created when the person engages in an IM. Example in point: the fact that something like `im.created` and `im.open` exists. I think you should be using one of those two events-- as to which, test and find out :^)
2) We shouldn't be reloading _everyones_ IMs after just one person joins the team. Instead, add a `_handle_blah` function that individually adds the new IM channel to the channels container. likewise when `im.close` is called.

Also, move this line to underneath the other handlers so its neater :^)

I also don't know if this is the right place/way to do this, but I'll let @CameronAavik or @TRManderson work that out :^)

uqcsbot/api.py
```diff
@@ -208,6 +208,16 @@ def _add_channel(self, chan_dict: dict) -> Channel:
...
         return chan
 
+    def reload_ims(self, *args):
+        for page in self._bot.api.im.list.paginate():
+            for im in page['ims']:
+                if im['is_user_deleted']:
+                    continue
+
+                im['is_private'] = True
+                im['name'] = im['id']
```
mitchmcdee: can probably omit this name altogether (see: https://api.slack.com/types/channel at about half way down where they talk about name being useless)

uqcsbot/api.py
```diff
@@ -208,6 +208,16 @@ def _add_channel(self, chan_dict: dict) -> Channel:
         self._channels_by_id[chan.id] = chan
         return chan
 
+    def reload_ims(self, *args):
```
mitchmcdee: provide a docstring saying what this function does and what an `im` is-- its not immediately clear :^)

mitchmcdee: also, given my other comment, I think this function can just be moved inline to where channels are added.

uqcsbot/api.py
```diff
@@ -219,6 +229,9 @@ def _initialise(self):
             for page in self._bot.api.channels.list.paginate():
                 for chan in page['channels']:
                     self._add_channel(chan)
+
+            self.reload_ims()
```
mitchmcdee: As per my other comment, add the code here instead ^^

uqcsbot/base.py
Prometheus9103: Yeah I completely agree it was just from my testing that im_open didn't. im_created does though, not sure how I missed that... Makes it much easier.

It was only there because channels is declared after those lines.

uqcsbot/api.py
```diff
@@ -230,7 +220,15 @@ def _initialise(self):
...
-            self.reload_ims()
+            # The ims cover the direct messages between the bot and users
+            for page in self._bot.api.im.list.paginate():
+                for im in page['ims']:
+                    if im['is_user_deleted']:
+                        continue
+
+                    im['name'] = im['id']
+                    im['is_private'] = True
+                    self._add_channel(im)
```
CameronAavik: I'm not sure if there is much benefit to pre-loading all the IMs. There are a lot of people in the slack. Why not just cache on the first im (which it looks like you have done with `_on_im_created`)?

Prometheus9103: Not sure I understand. The im_created will only trigger for new users joining the chat as that im is automatically made. The initialise is just there to capture the ims of pre existing users. I'm probably just misunderstanding something.

CameronAavik: Ah, so the ims list will only return people who have an im, not every person in the slack? If so then I'm fine with it.

Prometheus9103: It only returns all ims that the bot has access to

Prometheus9103: But I guess technically that is everyone in the slack? From my testing everyone gets an im channel with the bot. Unless that's different for larger slacks.

mitchmcdee: that's only slackbot, uqcsbot will be treated as any other user and require manual creation

mitchmcdee: afaik

Prometheus9103: Hmm okay, whenever I created a new user to test they automatically got  channel but they also got a channel with every other user so I guess that must just be low member count channels

uqcsbot/api.py
```diff
@@ -244,6 +255,13 @@ def get(self, name_or_id: str, default: T = None, use_cache: bool =True) -> Unio
     def __iter__(self):
         return iter(self._channels_by_id.values())
 
+    def _on_im_created(self, evt):
+        chan = evt['channel']
+
+        chan['name'] = chan['id']
+        chan['is_private'] = True
```
mitchmcdee: At this stage I don't think it's necessary to set `is_private` to true, even if it is. We should leave it up to slack events to tell us if its private. Ultimately, if we want to implement a script that needs to decide if a channel is private or not, we can just do something like `if is_group or is_im:`. Look at my #261 PR and implement an `is_im` field similar to the new `is_group` :^)

mubiquity:
From my testing im_close event doesn't seem to be triggered when a user closes the message

uqcsbot/api.py
mubiquity: Yeah that sounds reasonable to me but I didn't want to change the channels or remove channel.name etc in case something relied on it

mitchmcdee:
sgtm

uqcsbot/api.py
```diff
@@ -226,7 +229,6 @@ def _initialise(self):
                     if im['is_user_deleted']:
                         continue
 
-                    im['name'] = im['id']
                     im['is_private'] = True
```
mitchmcdee: why was name removed here? is name available in im.list? or were you meant to remove is_private? (as you did in the lower change)

mubiquity: Yeah that was a mistake, thanks

uqcsbot/api.py
```diff
@@ -156,6 +157,7 @@ def __init__(
         self.id = channel_id
         self.name = name
         self._member_ids = None
+        self._is_im = is_im
```
TRManderson: Leading underscore is a mismatch with everywhere else, including your type sig

uqcsbot/api.py
```diff
@@ -219,6 +222,16 @@ def _initialise(self):
             for page in self._bot.api.channels.list.paginate():
                 for chan in page['channels']:
                     self._add_channel(chan)
+
+            # The ims cover the direct messages between the bot and users
+            for page in self._bot.api.im.list.paginate():
+                for im in page['ims']:
+                    if im['is_user_deleted']:
```
TRManderson: Is `is_user_deleted` guaranteed to be in the response?

mubiquity: I believe so but I can replace it with a get if you'd like?

TRManderson: Nah this is fine

PR #260, opened 2018-03-08
Added !mock, a script that will mock previous messages in a channel

mitchmcdee:
image


uqcsbot/scripts/mock.py
```diff
@@ -0,0 +1,46 @@
...
+    Given a channel and a message index, will find the message at that index;
+    where messages are ordered from most recent to least recent. If the index is
+    greater than the number of messages in the channel, will return None.
+    '''
+    message_count = 0
+    # Add 1 to message_index as the 'limit' field is 1-indexed. For example, the
+    # 0th message (i.e the first message) can be retrieved by limiting the
+    # number of messages to 1.
+    message_index += 1
+    for page in bot.api.conversations.history.paginate(channel=channel_id, limit=message_index):
```
CameronAavik: Can we put an upper bound on the message_index so people don't put 1000 and try and break the API?

mitchmcdee: limited to 100 now :^) should be upper limit of 1 page i.e. 1 API call

mitchmcdee:
PTAL :^)

gricey432:
This would make a good little code golf puzzle

```
from random import choice
''.join(choice([c.upper, c.lower])() for c in message)
```

PR #263, opened 2018-03-08
Added !wolfram command

mubiquity:
A single command !wolfram that does two things. Depending on the --full arg

!wolfram --full 2x+8
!wolfram 2x+8 --full
Examples:
wolframfull

!wolfram on the other hand returns just a short text response that can usually be replied to. For example:

wolfram1
wolfram2
wolfram3
wolfram4

If Wolfram cannot come up with a conversation for that question (seems to be a rare case) then a short answer is instead supplied using their short answers API. 

The app id you get from wolfram is free for less than 2000 requests a month which I doubt would ever be exceeded.

### DIFF
--- a/uqcsbot/scripts/wolfram.py
+++ b/uqcsbot/scripts/wolfram.py
@@ -62,7 +62,6 @@ async def handle_wolfram(command: Command):
         error = json_response['error']
         if error == "No result is available":
             # If no conversational result is available just return a normal short answer
-            # TODO: Should the user be told they can't reply to the result?
             await short_answer(command)
             return
         else:
@@ -70,35 +69,85 @@ async def handle_wolfram(command: Command):
             return
 
     result = json_response['result']
+    reply_host = json_response['host']
     conversation_id = json_response['conversationID']
+    s_output = json_response.get('s', None)
 
     # TODO: Is there a better option than storing the id in the fallback?
-    # TODO: The footer is actually used in a check in handle_reply. This is probably fine unless someone else uses this footer but why would they? In any case a better solution would be good. Another non visible field would work.
+    # Here we store the conversation ID in the fallback so we can get it back later. We also store and indicator of this
     attachments = [{
-        'fallback': conversation_id,
+        'fallback': f'WolframCanReply {reply_host} {s_output} {conversation_id}',
         'footer': 'Further questions may be asked',
-        'text': result
+        'text': result,
     }]
 
     bot.post_message(command.channel, "", attachments=attachments)
 
+# TODO: Add 'ok' checks
 @bot.on('message')
-def handle_reply(evt: dict):
-    channel = evt['channel']
-
-    # If the message is not in a thread ignore it
-    if 'thread_ts' not in evt:
+async def handle_reply(evt: dict):
+    # If the message isn't from a thread or is from a bot ignore it (avoid those infinite loops)
+    if 'thread_ts' not in evt or ('subtype' in evt and evt['subtype'] == 'bot_message'):
         return
 
+    channel = evt['channel']
+
     thread_ts = evt['thread_ts']
-    thread_parent = bot.api.conversations.history(channel=channel, limit=1, inclusive=True, latest=thread_ts).paginate()
-    thread_parent = bot.api.conversations.history(limit=1).paginate()
-    bot.post_message(channel, thread_parent)
+    thread_parent = bot.api.conversations.history(channel=channel, limit=1, inclusive=True, latest=thread_ts)
 
+    if not thread_parent['ok']:
+        bot.post_message(channel, 'Sorry, something went wrong with slack', thread_ts=thread_ts)
+        return
 
-# @bot.on('message')
-# def temp(evt: dict):
-#     print(evt)
+    parent_message = thread_parent['messages'][0]
+    # If the threads parent wasn't by a bot ignore
+    if 'subtype' not in parent_message or parent_message['subtype'] != 'bot_message':
+        return
+
+    # Finally, we have to check that this is a Wolfram replyable message
+    # It is rare we would reach this point and not pass as who replies to a bot in a thread for another reason
+    parent_attachments = parent_message['attachments'][0]
+    parent_fallback = parent_attachments['fallback']
+    if 'WolframCanReply' not in parent_fallback:
+        return
+
+    # Now we can grab the conversation_id from the message and get the new question (s only sometimes appears)
+    _, reply_host, s_output, conversation_id = parent_fallback.split(' ')
+    new_question = evt['text']
+    s_output = '' if s_output == 'None' else s_output
+
+    # Slack annoyingly formats the reply_host link so we have to extract what we want:
+    reply_host = reply_host[1:-1].split('|')[0]
+
+    # Now we can ask Wolfram for the next answer:
+    api_url = f'{reply_host}/api/v1/conversation.jsp?'
+    params = {'appid': APP_ID, 'i': new_question, 'conversationid': conversation_id, 's': s_output}
+    http_response = await bot.run_async(requests.get, api_url, params=params)
+
+    if http_response.status_code != requests.codes.ok:
+        bot.post_message(channel, "There was a problem getting the response", thread_ts=thread_ts)
+        return
+
+    # Convert to json and check for an error
+    json_response = json.loads(http_response.content)
+    if 'error' in json_response:
+        bot.post_message(channel, json_response['error'], thread_ts=thread_ts)
+        return
+
+    # Otherwise grab the new stuff and post the reply.
+    reply = json_response['result']
+    conversation_id = json_response['conversationID']
+    reply_host = json_response['host']
+    s_output = json_response.get('s', None)
+
+    bot.post_message(channel, reply, thread_ts=thread_ts)
+
+    # Update the old conversation_id to indicate the new state
+    updated_attachments = parent_attachments.copy()
+    updated_attachments['fallback'] = f'WolframCanReply {reply_host} {s_output} {conversation_id}'
+
+    bot.api.chat.update(channel=channel, attachments=[updated_attachments], ts=thread_ts)
+
 
 async def short_answer(command: Command):
     """
--- a/uqcsbot/scripts/wolfram.py
+++ b/uqcsbot/scripts/wolfram.py
@@ -5,7 +5,6 @@ import json
 
 APP_ID = "UVKV2V-5XW2TETT69"
 
-# IDEA: Show the assumptions made?
 # TODO: Better naming of commands
 @bot.on_command("wolframfull")
 async def handle_wolframfull(command: Command):
@@ -79,10 +78,7 @@ async def handle_wolfram(command: Command):
             return
 
     # Unpack the response in order to construct the reply message
-    result = wolfram_result['result']  # This is the answer to our question
-    reply_host = wolfram_result['host']  # This is the hostname to ask the next question to
-    conversation_id = wolfram_result['conversationID']  # Used to continue the conversation
-    s_output = wolfram_result.get('s', None)  # s is only sometimes returned but is vital if it is returned
+    result, conversation_id, reply_host, s_output = extract_reply(wolfram_result)
 
     # TODO: Is there a better option than storing the id in the fallback?
     # Here we store the conversation ID in the fallback so we can get it back later.
@@ -110,7 +106,8 @@ async def handle_reply(evt: dict):
         bot.post_message(channel, 'Sorry, something went wrong with slack', thread_ts=thread_ts)
         return
 
-    parent_message = thread_parent['messages'][0]  # Limit=1 was used so the first (and only) message is what we want
+    # Limit=1 was used so the first (and only) message is what we want
+    parent_message = thread_parent['messages'][0]
     # If the threads parent wasn't by a bot ignore it
     if parent_message.get('subtype') != 'bot_message':
         return
@@ -126,9 +123,10 @@ async def handle_reply(evt: dict):
     # Recall the format of the fallback "identifier hostname s_output conversationID"
     _, reply_host, s_output, conversation_id = parent_fallback.split(' ')
     new_question = evt['text']  # This is the value of the message that triggered this whole response
-    s_output = '' if s_output == 'None' else s_output
+    s_output = '' if s_output == 'null' else s_output
 
     # Slack annoyingly formats the reply_host link so we have to extract what we want:
+    # The format is <http://www.domain.com|www.domain.com>
     reply_host = reply_host[1:-1].split('|')[0]
 
     # Now we can ask Wolfram for the next answer:
@@ -147,10 +145,7 @@ async def handle_reply(evt: dict):
         return
 
     # Otherwise grab the new stuff and post the reply.
-    reply = wolfram_answer['result']
-    conversation_id = wolfram_answer['conversationID']
-    reply_host = wolfram_answer['host']
-    s_output = wolfram_answer.get('s', None)
+    reply, conversation_id, reply_host, s_output = extract_reply(wolfram_answer)
 
     bot.post_message(channel, reply, thread_ts=thread_ts)
 
@@ -178,6 +173,16 @@ async def short_answer(search_query: str):
 
     return http_response.content
 
+def extract_reply(wolfram_response: dict) -> Tuple[str, str, str, str]:
+    """
+    Takes the response from the conversations API and returns it as a tuple containing the reply, conversation id,
+    reply host and s parameters. In that order.
+    """
+
+    return (wolfram_response['result'], # This is the answer to our question
+            wolfram_response['conversationId'], # Used to continue the conversation
+            wolfram_response['host'], # This is the hostname to ask the next question to
+            wolfram_response.get('s', 'null')) # s is only sometimes returned but is vital if it is returned
 
 def get_subpods(pods: list) -> Iterable[Tuple[str, dict]]:
     """
--- a/uqcsbot/scripts/wolfram.py
+++ b/uqcsbot/scripts/wolfram.py
@@ -5,6 +5,7 @@ import json
 
 APP_ID = "UVKV2V-5XW2TETT69"
 
+
 # TODO: Show the assumptions made?
 # TODO: Better naming of commands
 @bot.on_command("wolframfull")
@@ -58,6 +59,7 @@ async def handle_wolfram(command: Command):
         return
 
     bot.post_message(command.channel, http_response.content)
+
 
 # TODO: Should this really be a generator?
 def get_subpods(pods: list) -> Iterable[Tuple[str, dict]]:
--- a/uqcsbot/scripts/wolfram.py
+++ b/uqcsbot/scripts/wolfram.py
@@ -45,7 +45,7 @@ async def handle_wolframfull(command: Command):
 
 @bot.on_command("wolfram")
 async def handle_wolfram(command: Command):
-    """This uses wolframs short answers api to just return a simple short plaintext response"""
+    """This uses wolfram's conversation api to return a short response that can be replied to in a thread"""
     api_url = r"http://api.wolframalpha.com/v1/conversation.jsp"
     search_query = command.arg
     http_response = await bot.run_async(requests.get, api_url, params={'input': search_query, 'appid': APP_ID})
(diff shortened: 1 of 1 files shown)


mitchmcdee:
This looks lit. I'll give a code review for later :^)

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+from typing import Iterable, Tuple
+import requests
+import json
+
+APP_ID = "UVKV2V-5XW2TETT69"
+
+# IDEA: Show the assumptions made?
+# TODO: Better naming of commands
+@bot.on_command("wolframfull")
+async def handle_wolframfull(command: Command):
```
mitchmcdee: maybe instead of a new command, you can handle an optional variable? looks like it would remove some code duplication as well.

i.e. !wolfram something and then !wolfram something --full

if this is too annoying, consider it an optional improvement :^)

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+    http_response = await bot.run_async(requests.get, api_url, params={'input': search_query, 'appid': APP_ID})
+
+    # Check if the response is ok
+    if http_response.status_code != requests.codes.ok:
+        bot.post_message(command.channel, "There was a problem getting the response")
+        return
+
+    # Get the result of the query and determine if wolfram succeeded in evaluating it
+    result = json.loads(http_response.content)['queryresult']
+    if not result['success'] or result["error"]:
```
mitchmcdee: why two options? shouldn't just one of them suffice?

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+        bot.post_message(command.channel, "Please rephrase your query. Wolfram could not compute.")
```
mitchmcdee: should  you be returning here?

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+    # Check if the response is ok. A status code of 501 signifies that no result could be found.
+    if http_response.status_code == 501:
+        return "No short answer available. Try !wolframfull"
+    elif http_response.status_code != requests.codes.ok:
+        return "There was a problem getting the response"
+
+    return http_response.content
+
+
+def get_subpods(pods: list) -> Iterable[Tuple[str, dict]]:
```
mitchmcdee: what the heck is a subpod? if its short for something, use the elaborated version in the function name. Also, add an explanation of what a subpod is in this description

mitchmcdee: Also, I'd move this function to top-- it's generally easier to see helper functions before getting to where they're used rather than having to dig down into the code

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+    """
+    Yields subpods in the order they should be displayed
+
+    Yield: (pod_or_subpod_title, subpod)
+    """
+    for pod in pods:
+        for subpod in pod["subpods"]:
+            # Use the pods title if the subpod doesn't have its own title (general case)
+            title = subpod['title'] if subpod['title'] else pod['title']
```
mitchmcdee: why not `title = subpod.get('title', pod['title'])`? Will pod always have a title?

mitchmcdee: Are you meant to be checking if `title` exists in subpod or if its non-empty? because I think your code would KeyError if you're checking if it exists

mitchmcdee: If you're checking its non-empty, then yeah my suggestion won't work hehe-- but I think in that case you should make your code more verbose i.e. `if len(subpod['title']) != 0`

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+
+    bot.post_message(channel, reply, thread_ts=thread_ts)
+
+    # Update the old fallback to reflect the new state of the conversation
+    parent_attachments['fallback'] = f'WolframCanReply {reply_host} {s_output} {conversation_id}'
+
+    bot.api.chat.update(channel=channel, attachments=[parent_attachments], ts=thread_ts)
+
+
+async def short_answer(search_query: str):
```
mitchmcdee: would prefer `get_short_answer` :)

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+    """
+    This uses wolfram's short answers api to just return a simple short plaintext response.
+
+    This is used if the conversation api fails to get a result (for instance !wolfram pineapple is not a great
+    conversation starter but may be interesting.
+    """
+    api_url = r"http://api.wolframalpha.com/v1/result?"
```
mitchmcdee: I notice you use v1 here and v2 somewhere else, any reason?

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+    if http_response.status_code != requests.codes.ok:
+        bot.post_message(command.channel, "There was a problem getting the response")
+        return
+
+    # Get the result of the query and determine if wolfram succeeded in evaluating it
+    result = json.loads(http_response.content)['queryresult']
+    if not result['success'] or result["error"]:
+        bot.post_message(command.channel, "Please rephrase your query. Wolfram could not compute.")
+
+    subpods = get_subpods(result['pods'])
```
mitchmcdee: this needs renaming/a descriptive comment but I'll reference it in the function comments :P

mitchmcdee: If you don't rename it, add comments here to be explicit I guess

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+    message = ""
+    for title, subpod in subpods:
```
mitchmcdee: I wouldn't bother extracting subpods into its own variable-- just inline the generator

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+        plaintext = subpod["plaintext"]
+
+        # Prefer a plain text representation to the image
+        if plaintext != "" and plaintext != "* * * * * *":
```
mitchmcdee: why are these necessary? can you link an API query example?

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+            message += f'{title}: {plaintext}\n'
+        else:
+            image_url = subpod['img']['src']
+            image_title = subpod['img']['title']
+            message += f'{image_title}:\n{image_url}\n' if image_title else f'{image_url}\n'
```
mitchmcdee: I'd prefer `len(image_title) != 0` over `image_title`, its not clear that you're checking non-emptiness

uqcsbot/scripts/wolfram.py
mitchmcdee: lmk if you think they're sufficiently different responses that it warrants splitting

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+    This uses wolfram's conversation api to return a short response that can be replied to in a thread.
+    If the response cannot be replied to a general short answer response is displayed instead.
+
+    Example Usage:
+    !wolfram Solve Newton's Second Law for mass
+    !wolfram What is the distance from Earth to Mars?
+
+    and then start a thread to continue to conversation
+    """
+    api_url = r"http://api.wolframalpha.com/v1/conversation.jsp"
```
mitchmcdee: why `r""`?

uqcsbot/scripts/wolfram.py
mitchmcdee: and why `r""`?

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+# TODO: Better naming of commands
+@bot.on_command("wolframfull")
+async def handle_wolframfull(command: Command):
+    """
+    This posts the full results from wolfram query. Images and all
+
+    Example usage:
+    !wolframfull y = 2x + c
+    """
+    api_url = r"http://api.wolframalpha.com/v2/query?&output=json"
```
mitchmcdee: why `r""`?

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+        bot.post_message(command.channel, "There was a problem getting the response")
+        return
+
+    wolfram_result = json.loads(http_response.content)
+    if 'error' in wolfram_result:
+        error = wolfram_result['error']
+        if error == "No result is available":
+            # If no conversational result is available just return a normal short answer
+            short_response = await short_answer(search_query)
+            bot.post_message(command.channel, short_response)
```
mitchmcdee: I think you can remove the `await` here and inline the function call

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+            return
+        else:
+            bot.post_message(command.channel, error)
+            return
+
+    # Unpack the response in order to construct the reply message
+    result = wolfram_result['result']  # This is the answer to our question
+    reply_host = wolfram_result['host']  # This is the hostname to ask the next question to
+    conversation_id = wolfram_result['conversationID']  # Used to continue the conversation
+    s_output = wolfram_result.get('s', None)  # s is only sometimes returned but is vital if it is returned
```
mitchmcdee: if its vital, explain why and what it is! :P

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+
+    # TODO: Is there a better option than storing the id in the fallback?
+    # Here we store the conversation ID in the fallback so we can get it back later.
+    # We also store an identifier string to check against later and the reply_host and s_output string
+    attachments = [{
+        'fallback': f'WolframCanReply {reply_host} {s_output} {conversation_id}',
```
mitchmcdee: is `s_output` is None, this will have a double space, is that an issue?

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+    # Unpack the response in order to construct the reply message
+    result = wolfram_result['result']  # This is the answer to our question
+    reply_host = wolfram_result['host']  # This is the hostname to ask the next question to
+    conversation_id = wolfram_result['conversationID']  # Used to continue the conversation
+    s_output = wolfram_result.get('s', None)  # s is only sometimes returned but is vital if it is returned
+
+    # TODO: Is there a better option than storing the id in the fallback?
+    # Here we store the conversation ID in the fallback so we can get it back later.
+    # We also store an identifier string to check against later and the reply_host and s_output string
+    attachments = [{
```
mitchmcdee: explain what's the attachment used for and why

uqcsbot/scripts/wolfram.py
mitchmcdee: also I'm happy to keep it for now, but I'm sure we can do something better than stuffing it in fallback hehe :^) maybe we can build another robot brain (essentially just an in-memory key-value store we used in our jsbot)

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+        'fallback': f'WolframCanReply {reply_host} {s_output} {conversation_id}',
+        'footer': 'Further questions may be asked',
+        'text': result,
+    }]
+
+    bot.post_message(command.channel, "", attachments=attachments)
+
+
+@bot.on('message')
+async def handle_reply(evt: dict):
```
mitchmcdee: Add a function docstring here :)

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+    # If the message isn't from a thread or is from a bot ignore it (avoid those infinite loops)
+    if 'thread_ts' not in evt or (evt.get('subtype') == 'bot_message'):
```
mitchmcdee: can remove redundant parentheses

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+        return
+
+    channel = evt['channel']
+    thread_ts = evt['thread_ts']  # This refers to time the original message
+    thread_parent = bot.api.conversations.history(channel=channel, limit=1, inclusive=True, latest=thread_ts)
+
+    if not thread_parent['ok']:
+        bot.post_message(channel, 'Sorry, something went wrong with slack', thread_ts=thread_ts)
```
mitchmcdee: so what does this section do? does it post to the thread? doesn't being able to post to the thread assume that there's a thread parent, which we just learnt wasn't ok?

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+        return
+
+    parent_message = thread_parent['messages'][0]  # Limit=1 was used so the first (and only) message is what we want
```
mitchmcdee: put this comment above the line

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+    parent_attachments = parent_message['attachments'][0]  # The message is formatted with only one attachment
+    parent_fallback = parent_attachments['fallback']
+    if 'WolframCanReply' not in parent_fallback:
+        return
+
+    # Now we can grab the conversation_id from the message and get the new question (s only sometimes appears)
+    # Recall the format of the fallback "identifier hostname s_output conversationID"
+    _, reply_host, s_output, conversation_id = parent_fallback.split(' ')
+    new_question = evt['text']  # This is the value of the message that triggered this whole response
+    s_output = '' if s_output == 'None' else s_output
```
mitchmcdee: why `'None'`? was the `s_ouput` from when it was set above meant to set it to the string `None` if there wasn't an s_output? if so, I don't think you made it a string-- also, if that was your intention, I would call it 'null' to differentiate between `None`

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+
+    # Slack annoyingly formats the reply_host link so we have to extract what we want:
+    reply_host = reply_host[1:-1].split('|')[0]
```
mitchmcdee: State how Slack formats it so its clear to anyone reading this without having to test it themselves. Provide an example if need be

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+
+    # Now we can ask Wolfram for the next answer:
```
mitchmcdee: I'm seeing a lot of this boilerplate code for making a conversation api call, checking its ok and then extracting the answer-- looks like it can be abstracted into its own function and clear a bit up maybe?

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+    wolfram_answer = json.loads(http_response.content)
+    if 'error' in wolfram_answer:
+        bot.post_message(channel, wolfram_answer['error'], thread_ts=thread_ts)
+        return
+
+    # Otherwise grab the new stuff and post the reply.
+    reply = wolfram_answer['result']
+    conversation_id = wolfram_answer['conversationID']
+    reply_host = wolfram_answer['host']
+    s_output = wolfram_answer.get('s', None)
```
mitchmcdee: this section is all duplicated code-- abstract the functionality :^)

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
+from uqcsbot import bot, Command
+from typing import Iterable, Tuple
+import requests
+import json
+
+APP_ID = "UVKV2V-5XW2TETT69"
+
+# IDEA: Show the assumptions made?
```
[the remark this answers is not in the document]
mubiquity: The api returns the assumptions it made during the processing. I wasn't sure whether or not to include it. Leaning towards not as the results can be long enough as is and are usually on point.

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
+from uqcsbot import bot, Command
+from typing import Iterable, Tuple
+import requests
+import json
+
+APP_ID = "UVKV2V-5XW2TETT69"
```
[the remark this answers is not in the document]
mubiquity: Yeah I will but I wasn't sure whether or not you guys would want to replace it with another id

uqcsbot/scripts/wolfram.py
mubiquity: I was on the fence about this too. I was concerned that the usual  "- - cmd" would interfere with math input but mainly that they do do pretty different things.

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,192 @@
...
+
+# IDEA: Show the assumptions made?
+# TODO: Better naming of commands
```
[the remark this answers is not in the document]
mubiquity: Suggestions welcome :)

uqcsbot/scripts/wolfram.py
mubiquity: Just a url habit of mine sometimes.

uqcsbot/scripts/wolfram.py
mubiquity: Is this about the success/error? They mean slightly different things and I think you can have one fine and the other not fine.

uqcsbot/scripts/wolfram.py
mubiquity: I figured it would convert to a string but null is better anyway for clarity sake.

mubiquity:
Is there a reason I can't seem to see all the comments you left aside from in my email? I only just noticed there are way more.

uqcsbot/scripts/wolfram.py
mubiquity: Often the plaintext will just be a bunch of stars. From my testing 6 of them was the most common. But in trying to find you an example I found results that had more than that. I can't find a reference in the documentation anywhere.

uqcsbot/scripts/wolfram.py
mubiquity: I would if I knew but the docs only say "This parameter should only be used in the rare case that your result includes the s output parameter" and "When present, this value must be propagated to the next query to ensure continuity of conversations". They don't even explain what it stands for.

uqcsbot/scripts/wolfram.py
mubiquity: This is just a check to ensure that the thread parent was returned. I was running into a lot of auth issues in my test slack so if we see this come up that is the likely culprit.

mitchmcdee:
I personally never use the emails for anything more than a ping + link to the github bahahhaa

mitchmcdee:
lmk when you want another code review (not sure if you're still changing stuff)

mubiquity:
I'll happily change more. So just whenever you have time :)

uqcsbot/scripts/wolfram.py
mitchmcdee: I would remove the r unless its needed :^) just to be that little bit extra clear

uqcsbot/scripts/wolfram.py
mitchmcdee: sgtm

uqcsbot/scripts/wolfram.py
mitchmcdee: I'd upgrade to latest API if it doesn't change anything, just for consistency and longevity :^)

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,226 @@
...
+        # Update the old fallback to reflect the new state of the conversation
+        parent_attachments['fallback'] = f'WolframCanReply {reply_host} {s_output} {conversation_id}'
+
+        bot.api.chat.update(channel=channel, attachments=[parent_attachments], ts=thread_ts)
+
+
+
+
+
+
```
mitchmcdee: remove all these newlines

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,226 @@
...
+from typing import Iterable, Tuple, Optional
+from base64 import b64decode
+import requests
+import json
+
+APP_ID = b64decode('RzU0S1VBLVVHWTdHR0hWUlg=').decode('utf-8')
+
+
+
+
```
mitchmcdee: remove these unnecessary new lines

mubiquity:
The conversation api doesn't have  a v2 but the short answer one does so I have changed that.

uqcsbot/scripts/wolfram.py
```diff
@@ -0,0 +1,218 @@
+from uqcsbot import bot, Command
+from typing import Iterable, Tuple, Optional
+from base64 import b64decode
+import requests
+import json
+
+APP_ID = b64decode('RzU0S1VBLVVHWTdHR0hWUlg=').decode('utf-8')
```
CameronAavik: For the jsbot we have used an environment variable for this instead of hardcoding it. Would we be able to load this from an environment variable instead? (with a link on how to get an App ID)

mitchmcdee:
In future I'd suggest splitting this PR into two halves (!wolfram and !wolframfull) though so that it can be merged quicker, good job though keep it up :^)

mubiquity:
...I did just change it to an arg based system though hahaha

mitchmcdee:
bahahaha no wozzas still lgtm. also, really good use of comments throughout your code 👍 makes it much easier to understand

mitchmcdee:
is the code ready to submit? (i.e. tested and still works?)

mubiquity:
Yep, all tested and no issues. Though it only works with the changes from #261. I completely forgot until I had to test just then and had to re-pull the changes. I think it is just some auth issue with the test slack I use which the uqcstesting slack doesn't have

mitchmcdee:
interesting. I'll merge anyway as the code doesn't depend on #261 and the bot isn't currently in use anywho

mitchmcdee:
oh looooooooool sorry @CameronAavik for some reason I thought this one was LGTM'd by you bahahaha

PR #265, opened 2018-03-09
Added !pastexams

mubiquity:
Just like the old !pastexams. Returns a list of past exams, uses the channel name if not supplied with a course code.

pastexams1


uqcsbot/scripts/pastexams.py
```diff
@@ -0,0 +1,75 @@
...
+    Usage:
+    !pastexams CSSE2310
+    or if in a course channel:
+    !pastexams
+    """
+    course_code = command.arg if command.has_arg() else command.channel.name
+    past_exams = await get_past_exams(course_code)
+
+    # We use attachments to improve the formatting
+    attachments = [
```
mitchmcdee: just inline this below :^)

Issue #254, opened 2018-03-05
Add testing framework


gricey432:
Each command should provide automated tests to allow commands to be tested without a slack workspace

mitchmcdee:
yyyyyyESSsss we very much need tests bahahahaha

gricey432:
Framework is in. Now we need more tests

mitchmcdee:
Added a work in progress version with async support in #277

mitchmcdee:
Will close this issue once #292 is complete (and subsequently will remove #277)

mitchmcdee:
Adding of tests has been moved to #294

mitchmcdee:
as the sync version is complete, and as the process of porting the useful sections of #277 has been moved to #296, I'll close this issue now :^)

PR #266, opened 2018-03-12
Added a test framework

gricey432:
I was trying to get testing working with async but @TRManderson gave the impression that what I was trying to do isn't possible. So I've done this instead.

Not super happy with it, and very open to suggestions.

Also fixed a bunch of PEP8 stuff while I was there.

Progresses on #254


requirements.txt
```diff
@@ -1 +0,0 @@
--e .
```
gricey432: This causes issues with Pycharm's detection of requirements

TRManderson:
Just a general comment - this isn't Polymathian. UQCSBot doesn't necessarily follow the Polymathian style guide, and "Because PyCharm" isn't necessarily a good enough reason like it is at work.

mitchmcdee:
I'm all for the style guide changes, but it heavily pollutes the PR-- could it be separated into its own PR?

gricey432:
@TRManderson all of the style changes are PEP8 apart from the `requirements.txt`. I'd say PEP8 should apply everywhere.
As for the `requirements.txt` file, it was more a combo of "it messes with pycharm" and "there's no documented reason this needs to exist". Idno what it's for, but I can undo that part if it actually has a use.

@mitchmcdee how keen are you on that? I did all this in one commit, so I'd have to split the changes by hand which would suck a bit

mitchmcdee:
well they're just style changes with no dependencies-- should be easy to just copy paste each file into a new branch and then `git checkout -- <file>`. Would just make reviewing a heck of a lot easier :^)

gricey432:
@mitchmcdee after many screw ups with interactive rebasing, I've managed to split it into two branches. This one only has the test stuff now.

mitchmcdee:
kek 😂 coulda just done it in a new commit!

mitchmcdee:
Could you add a little blurb to the README explaining how to run tests?

gricey432:
Added readme

README.md
```diff
@@ -22,7 +22,7 @@ If a bot was available, it will now be running on uqcstesting.
 ## Running the bot in a custom Slack team
 
 1. [Create a Slack workspace](https://slack.com/create)
-2. Run `python3 setup.py install`
+2. Run `pip install -e .`
```
mitchmcdee: this is my bad for forgetting to remove step 2 when I re-added it, but you can probably omit this line entirely seeing as though its stated in 'Setting up the bot'

PR #267, opened 2018-03-12
Bot name printing on dev start.

tgrkzus:
Thought it might be helpful to know what bot you were assigned to.

Since I wanted to use the logger for this message I went ahead and did my best at changing all the errors on the bot assignment to use the logger. Not at all familiar with this so if I've done something wrong please do yell at me.

### DIFF
--- a/uqcsbot/__init__.py
+++ b/uqcsbot/__init__.py
@@ -137,7 +137,7 @@ def main():
         # If in development mode, attempt to allocate an available bot token,
         # else stick with the default. If no bot could be allocated, exit.
         bot_info = get_test_bot_token() if args.dev else SLACK_BOT_TOKEN
-        if bot_info is not None:
+        if bot_info is None:
             logger.error('Something went wrong during bot allocation. '
                   'Please ensure there are bots available and try again later. '
                   'Exiting.')
--- a/uqcsbot/__init__.py
+++ b/uqcsbot/__init__.py
@@ -24,6 +24,7 @@ SECRET_BOT_MEETING_ROOM = 'G9JJXHF7S'
 # Mitch's UQCSTesting Slack API Token. No touchie >:(
 API_TOKEN = b64decode('eG94cC0yNjA3ODI2NzQ2MTAtMjYwMzQ1MTQ0NTI5LTMyNTEyMzU5ODExNS01YjdmYjlhYzAyZWYzNDAyNTYyMTJmY2Q2YjQ1NmEyYg==').decode('utf-8')
 
+logger = logging.getLogger("uqcsbot")
 
 def is_active_bot(user_id):
     '''
@@ -60,24 +61,42 @@ def get_test_bot_token():
     '''
     Pings a channel on the UQCSTesting Slack that contains all the available
     bots, and Mitch. We can poll this channel to find  bots which are 'away'
-    (that is, not currently being used by anyone else) and return their
-    respective BOT_TOKEN.
+    (that is, not currently being used by anyone else)
+
+    Returns the bot user id and token in a dictionary: {'id': ..., 'token': ...}
     '''
     api_url = 'https://slack.com/api/conversations.members'
     response = requests.get(api_url, params={'token': API_TOKEN, 'channel': SECRET_BOT_MEETING_ROOM})
     if response.status_code != requests.codes['ok']:
-        print(f'Error: received status code {response.status.code}')
+        logger.error(f'Received status code {response.status.code}')
         sys.exit(1)
 
     json_contents = json.loads(response.content)
     if not json_contents['ok']:
-        print('Error: {0}'.format(json_contents['error']))
+        logger.error('{0}'.format(json_contents['error']))
         sys.exit(1)
 
     for user_id in json_contents['members']:
         if is_available_bot(user_id):
-            return BOT_TOKENS.get(user_id, None)
+            return {'id': user_id, 'token': BOT_TOKENS.get(user_id, None)}
     return None
+
+def get_test_bot_name(id):
+    '''
+    Get's a bot's name from their id
+    '''
+    api_url = 'https://slack.com/api/users.info'
+    response = requests.get(api_url, params={'token': API_TOKEN, 'user': id})
+    if response.status_code != requests.codes['ok']:
+        logger.error(f'Received status code {response.status.code}')
+        sys.exit(1)
+
+    json_contents = json.loads(response.content)
+    if not json_contents['ok']:
+        logger.error('{0}'.format(json_contents['error']))
+        sys.exit(1)
+
+    return json_contents['user']['name']
 
 
 def main():
@@ -117,13 +136,14 @@ def main():
     else:
         # If in development mode, attempt to allocate an available bot token,
         # else stick with the default. If no bot could be allocated, exit.
-        bot_token = get_test_bot_token() if args.dev else SLACK_BOT_TOKEN
-        if bot_token is None:
-            print('Something went wrong during bot allocation. '
+        bot_info = get_test_bot_token() if args.dev else SLACK_BOT_TOKEN
+        if bot_info is not None:
+            logger.error('Something went wrong during bot allocation. '
                   'Please ensure there are bots available and try again later. '
                   'Exiting.')
             sys.exit(1)
-        bot.run(bot_token, SLACK_VERIFICATION_TOKEN)
+        logger.info("Bot name: " + get_test_bot_name(bot_info['id']))
+        bot.run(bot_info['token'], SLACK_VERIFICATION_TOKEN)
 
 if __name__ == "__main__":
     main()
--- a/uqcsbot/__init__.py
+++ b/uqcsbot/__init__.py
@@ -95,12 +95,6 @@ def get_free_test_bot():
             return info
     return None
 
-def get_bot_token(id):
-    '''
-    Returns the bot token associated with a user id. None if this is not associated with a bot
-    '''
-    return BOT_TOKENS.get(id, None)
-
 
 def main():
     # Import scripts
--- a/uqcsbot/__init__.py
+++ b/uqcsbot/__init__.py
@@ -29,13 +29,21 @@ logger = logging.getLogger("uqcsbot")
 def get_bot_info(user_id):
     '''
     Returns info about the bot
+
+    See https://api.slack.com/methods/users.info for the contents of info
     '''
     api_url = 'https://slack.com/api/users.info'
     response = requests.get(api_url, params={'token': API_TOKEN, 'user': user_id})
+
     if response.status_code != requests.codes['ok']:
-        return False
+        logger.error(f'Received status code {response.status.code}')
+        sys.exit(1)
 
     json_contents = json.loads(response.content)
+    if not json_contents['ok']:
+        logger.error(json_contents['error'])
+        sys.exit(1)
+
     return json_contents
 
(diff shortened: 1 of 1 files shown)


uqcsbot/__init__.py
```diff
@@ -60,25 +61,43 @@ def get_test_bot_token():
...
     response = requests.get(api_url, params={'token': API_TOKEN, 'channel': SECRET_BOT_MEETING_ROOM})
     if response.status_code != requests.codes['ok']:
-        print(f'Error: received status code {response.status.code}')
+        logger.error(f'Received status code {response.status.code}')
         sys.exit(1)
 
     json_contents = json.loads(response.content)
     if not json_contents['ok']:
-        print('Error: {0}'.format(json_contents['error']))
+        logger.error('{0}'.format(json_contents['error']))
```
mitchmcdee: this can probably just be `logger.error(json_contents['error'])` now

uqcsbot/__init__.py
```diff
@@ -60,25 +61,43 @@ def get_test_bot_token():
     '''
```
mitchmcdee: I would rename this function as `get_test_bot()` instead now

uqcsbot/__init__.py
```diff
@@ -60,25 +61,43 @@ def get_test_bot_token():
...
+        logger.error('{0}'.format(json_contents['error']))
         sys.exit(1)
 
     for user_id in json_contents['members']:
         if is_available_bot(user_id):
-            return BOT_TOKENS.get(user_id, None)
+            return {'id': user_id, 'token': BOT_TOKENS.get(user_id, None)}
     return None
 
+def get_test_bot_name(id):
```
mitchmcdee: we already make an API call to `users.info`, maybe try restructuring the code to use that?

maybe something like `get_test_bot()` which calls `get_active_bot_name` which can return None if something bad occurred (i.e. bot was already taken)

uqcsbot/__init__.py
```diff
@@ -60,25 +61,43 @@ def get_test_bot_token():
...
+    api_url = 'https://slack.com/api/users.info'
+    response = requests.get(api_url, params={'token': API_TOKEN, 'user': id})
+    if response.status_code != requests.codes['ok']:
+        logger.error(f'Received status code {response.status.code}')
+        sys.exit(1)
+
+    json_contents = json.loads(response.content)
+    if not json_contents['ok']:
+        logger.error('{0}'.format(json_contents['error']))
+        sys.exit(1)
```
mitchmcdee: we should probably replace the lazy `sys.exit(1)` with proper exception throwing and then just handle it below when we first call `get_test_bot()`

uqcsbot/__init__.py
```diff
@@ -136,14 +131,14 @@ def main():
...
-        if bot_info is None:
+        test_bot = get_free_test_bot() if args.dev else SLACK_BOT_TOKEN
+        if test_bot is None:
             logger.error('Something went wrong during bot allocation. '
                   'Please ensure there are bots available and try again later. '
                   'Exiting.')
             sys.exit(1)
-        logger.info("Bot name: " + get_test_bot_name(bot_info['id']))
-        bot.run(bot_info['token'], SLACK_VERIFICATION_TOKEN)
+        logger.info("Bot name: " + get_test_bot_name(test_bot['info']))
```
mitchmcdee: won't this straight up break if an env var token is opted for?

uqcsbot/__init__.py
```diff
@@ -136,14 +131,14 @@ def main():
     else:
         # If in development mode, attempt to allocate an available bot token,
         # else stick with the default. If no bot could be allocated, exit.
-        bot_info = get_test_bot_token() if args.dev else SLACK_BOT_TOKEN
-        if bot_info is None:
+        test_bot = get_free_test_bot() if args.dev else SLACK_BOT_TOKEN
```
mitchmcdee: these are two different types, maybe split it up

mitchmcdee: i.e. if args.dev then get a test_bot, else just go through with the env var?

uqcsbot/__init__.py
```diff
@@ -26,27 +26,30 @@
 
 logger = logging.getLogger("uqcsbot")
 
-def is_active_bot(user_id):
+def get_bot_info(user_id):
```
mitchmcdee: seeing as though this doesn't check for whether its a bot at all, I'd fix the description and rename it `get_user_info`

uqcsbot/__init__.py
```diff
@@ -26,27 +26,30 @@
...
     '''
-    Returns true if the given user_id is an active bot (i.e. not deleted).
+    Returns info about the bot
     '''
     api_url = 'https://slack.com/api/users.info'
     response = requests.get(api_url, params={'token': API_TOKEN, 'user': user_id})
     if response.status_code != requests.codes['ok']:
         return False
 
     json_contents = json.loads(response.content)
```
mitchmcdee: can inline this now

uqcsbot/__init__.py
```diff
@@ -26,27 +26,30 @@
...
 
+def is_bot_active(info):
+    '''
+    Returns true if the provided bot info describes an active bot (i.e. not deleted)
+    '''
+    return info['ok'] and info['user']['is_bot'] and not info['user']['deleted']
 
-def is_available_bot(user_id):
+
+def is_bot_avaliable(info, user_id):
```
mitchmcdee: why do you pass it `info`?

uqcsbot/__init__.py
tgrkzus: No idea how to do exceptions in Python well. Is it overkill to write a custom exception type?

uqcsbot/__init__.py
```diff
@@ -26,27 +26,30 @@
...
     response = requests.get(api_url, params={'token': API_TOKEN, 'user': user_id})
     if response.status_code != requests.codes['ok']:
         return False
 
     json_contents = json.loads(response.content)
-    user = json_contents['user']
-    return json_contents['ok'] and user['is_bot'] and not user['deleted']
+    return json_contents
 
+def is_bot_active(info):
```
mitchmcdee: `bot_info`

uqcsbot/__init__.py
```diff
@@ -73,30 +78,20 @@ def get_test_bot_token():
...
-        logger.error('{0}'.format(json_contents['error']))
+        logger.error(json_contents['error'])
         sys.exit(1)
 
     for user_id in json_contents['members']:
-        if is_available_bot(user_id):
-            return {'id': user_id, 'token': BOT_TOKENS.get(user_id, None)}
+        info = get_bot_info(user_id)
+        if is_bot_active(info) and is_bot_avaliable(info, user_id):
+            return {'info': info, 'token': BOT_TOKENS.get(user_id, None)}
```
mitchmcdee: `users.info` returns the user_id, so we can probably just return the info here and get the BOT_TOKEN later. This avoids relying on an awkward dictionary

uqcsbot/__init__.py
```diff
@@ -73,30 +78,20 @@ def get_test_bot_token():
...
     return None
 
-def get_test_bot_name(id):
+def get_test_bot_name(info):
```
mitchmcdee: this function doesn't seem necessary

uqcsbot/__init__.py
```diff
@@ -117,13 +139,19 @@ def main():
...
+            test_bot = get_free_test_bot()
+            if test_bot is None:
+                logger.error('Something went wrong during bot allocation. '
+                      'Please ensure there are bots available and try again later. '
+                      'Exiting.')
+                sys.exit(1)
+            token = get_bot_token(test_bot['user']['id'])
+            logger.info("Bot name: " + test_bot['user']['name'])
+        else:
+            token = SLACK_BOT_TOKEN
```
mitchmcdee: can just move this to the top now and remove the `else`

uqcsbot/__init__.py
```diff
@@ -117,13 +139,19 @@ def main():
...
-            sys.exit(1)
-        bot.run(bot_token, SLACK_VERIFICATION_TOKEN)
+        if args.dev:
+            test_bot = get_free_test_bot()
+            if test_bot is None:
+                logger.error('Something went wrong during bot allocation. '
+                      'Please ensure there are bots available and try again later. '
+                      'Exiting.')
+                sys.exit(1)
+            token = get_bot_token(test_bot['user']['id'])
```
mitchmcdee: `get_bot_token`is simple enough to just inline it. If it looks too confusing, just pull out the id as its own variable
i.e.
```
bot_id = test_bot['user']['id']
bot_token = BOT_TOKENS.get(bot_id, None)
```

be sure to handle if bot_token is `None` too

uqcsbot/__init__.py
mitchmcdee: also maybe call it `bot_token` for clarity

uqcsbot/__init__.py
```diff
@@ -56,29 +68,39 @@ def is_available_bot(user_id):
...
 
     for user_id in json_contents['members']:
-        if is_available_bot(user_id):
-            return BOT_TOKENS.get(user_id, None)
+        info = get_user_info(user_id)
+        if is_bot_active(info) and is_bot_avaliable(user_id):
+            return info
     return None
 
+def get_bot_token(id):
```
mitchmcdee: as per below-- this function is simply enough to inline

uqcsbot/__init__.py
```diff
@@ -24,28 +24,40 @@
...
     json_contents = json.loads(response.content)
-    user = json_contents['user']
-    return json_contents['ok'] and user['is_bot'] and not user['deleted']
+    if not json_contents['ok']:
+        logger.error(json_contents['error'])
+        sys.exit(1)
+
+    return json_contents
+
+def is_bot_active(bot_info):
```
mitchmcdee: this function name implies you're checking if the bot is active, rather than if its an active bot. Rename to `is_active_bot` to be clear :^)

uqcsbot/__init__.py
```diff
@@ -24,28 +24,40 @@
...
+    '''
+    Returns true if the provided bot info describes an active bot (i.e. not deleted)
```
mitchmcdee: change to `provided info` as we're not certain its a bot yet

uqcsbot/__init__.py
```diff
@@ -24,28 +24,40 @@
...
+    '''
+    return bot_info['ok'] and bot_info['user']['is_bot'] and not bot_info['user']['deleted']
```
mitchmcdee: is `is_bot` present on non-bot users? I'd probably change it to `bot_info['user'].get('is_bot', False)` just to be safe

uqcsbot/__init__.py
mitchmcdee: Also, update the input param to reflect this (change to `user_info`)

uqcsbot/__init__.py
```diff
@@ -56,29 +68,39 @@ def is_available_bot(user_id):
...
 
     for user_id in json_contents['members']:
-        if is_available_bot(user_id):
-            return BOT_TOKENS.get(user_id, None)
+        info = get_user_info(user_id)
+        if is_active_bot(info) and is_bot_avaliable(user_id):
+            return info
     return None
 
+def get_bot_token(id):
```
mitchmcdee: forgot to remove this ;)

PR #268, opened 2018-03-12
Style changes

gricey432:
Fixed a couple of small PEP8 violations and removed the `requirements.txt` because it didn't seem to do anything and was messing with pycharm.

Split out of #266


uqcsbot/scripts/welcome.py
```diff
@@ -1,9 +1,11 @@
-# Welcomes new users to UQCS Slack and check for member milestones
+"""
+Welcomes new users to UQCS Slack and check for member milestones
+"""
 from uqcsbot import bot
 import asyncio
 
-MEMBER_MILESTONE = 50;  # Number of members between posting a celebration
-MESSAGE_PAUSE = 2.5;   # Number of seconds between sending bot messages
```
mitchmcdee: lol

requirements.txt
```diff
@@ -1 +0,0 @@
--e .
```
mitchmcdee: Does this affect installation at all? and if that's the only line, should we just delete the file?

CameronAavik: I remember tom telling me this was necessary to keep, but I can't remember why

gricey432: I removed it and it installed just fine for me, I've never seen anything like this before. @TRManderson ?

TRManderson: The answer is "it depends".

@gricey432, you're very used to doing `pip install -e .` from work stuff, but that's not the normal workflow for most python devs.

people using `pipenv` need the `requirements.txt`
people using `requirements.txt` need it
people using `setup.py` are covered
people using `pip install .` are covered

gricey432: But those first two methods won't work with tests_requires right? So having it there is a bit confusing.

gricey432: The `README.md` has been changed to tell people to use `pip`, so I'd say we don't need this anymore. Keeping it as simple as possible for a first-year to look around the repo and figure out what stuff is for

PR #274, opened 2018-04-13
Added define command

mubiquity:
Adds a command that returns the dictionary definition of the argument.

define

### DIFF
--- a/uqcsbot/scripts/define.py
+++ b/uqcsbot/scripts/define.py
@@ -1,0 +1,37 @@
+from uqcsbot import bot, Command
+import requests
+import json
+
+API_URL = "http://api.pearson.com/v2/dictionaries/laad3/entries?limit=1"
+
+@bot.on_command("define")
+async def define(command: Command):
+    '''
+    `!define <TEXT>` - Gets the dictionary definition of TEXT
+    '''
+    query = command.arg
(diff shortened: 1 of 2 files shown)


mitchmcdee:
isn't adding it as an attachment pretty much equivalent to using `>>>`? Also, would be great if you could include some tests (as per gricey's new test framework)

mubiquity:
Using an attachment adds the grey bar which I thought was a nicer format for the response. Can change if you disagree.

Will do, forgot that was a thing now. But I will wait for another script to implement them first so I can get some idea of what is wanted.

mitchmcdee:
image

`>>>` does that too though-- and would just make for more obvious code intention imo

as for tests, sgtm-- i'm busy for a while with uni but if you want to give it a crack i'm pretty sure gricey submitted a few examples of testing base, maybe go off that?

mubiquity:
Oh, my bad. I see what you mean. Didn't know that piece of slack formatting.

mubiquity:
I tried to make some tests but apparently they aren't compatible with async?

mitchmcdee:
yeah I wrote async fixes but haven't been bothered cleaning the code yet for a PR, I'll try take a look after wednesday-- if you wanna take a look though it should be in my forked repo

PR #276, opened 2018-05-31
Added rate limiting to api calls

### LINKED ISSUE #249: Handle Rate Limiting of Slack API
It seems that the issues we are having are caused by rate limiting in the Slack Web API: https://api.slack.com/docs/rate-limits

The reason we only see this on our actual slack and not the test slack is because there are so many members that it paginates the results when looping through the channel members, and we are not rate limiting our pagination attempts.

We apparently need to listen for a HTTP 429 Too Many Requests response and that will tell us how long to wait (in seconds!) until we should try again.

mitchmcdee:
As per title, adds rate limit handling to api calls.
Comments and feedback appreciated :^)

Fixes #249.

### DIFF
--- a/uqcsbot/api.py
+++ b/uqcsbot/api.py
@@ -1,4 +1,5 @@
 from functools import partial
+import time
 from slackclient import SlackClient
 import asyncio
 import threading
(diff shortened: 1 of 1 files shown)


uqcsbot/api.py
```diff
@@ -67,17 +68,30 @@ def __call__(self, **kwargs) -> dict:
...
-            loop = asyncio.get_event_loop()
-            return loop.run_in_executor(None, fn)
+        retry_count = 0
+        while retry_count < 5:
+            if self._async:
+                loop = asyncio.get_event_loop()
+                result = loop.run_in_executor(None, fn)
+            else:
+                result = fn()
+            if not result['ok'] and result['error'] == 'ratelimited':
```
TRManderson: Can't just use result like this - needs to be awaited if `self._async`

mitchmcdee: like, `result = await loop.run_in_executor(None, fn)`?

mitchmcdee: like, `result = await loop.run_in_executor(None, fn)`?

mitchmcdee:
re: `Try doing an inner async function that does the looping and passing that in to the loop.run_in_executor call`

for what reason do I need to do that? Isn't that going to look super duplicate-y if I need to do the loops for non-async too? or what did you have in mind here?

mitchmcdee:
@TRManderson PTAL <3

TRManderson:
Added a PR that builds on this. Yes it looks hella duplicate-y, but unfortunately that's what you get in Python.

PR #278, opened 2018-06-05
Async Rate Limiting

TRManderson:
Uses #276 as a base, but adds more sophisticated async logic.

We could totally cheat and just use the sync method in a call to `loop.run_in_executor` but this should theoretically play nicer with our event loop?

If we'd prefer less duplication over incredibly slight perf benefits, I'll rewrite.

### DIFF
--- a/uqcsbot/api.py
+++ b/uqcsbot/api.py
@@ -91,9 +91,9 @@ class APIMethodProxy(object):
             return result
         if self._is_async:
             loop = asyncio.get_event_loop()
-            loop.run_in_executor(None, call_inner)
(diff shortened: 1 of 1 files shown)


mitchmcdee:
Would defs prefer a less duplicated version just so that we can keep core code as readable as possible

TRManderson:
Less duplicated version pushed

mitchmcdee:
Much better <3 does it work though? I notice no return for async?

TRManderson:
lel I haven't tested, will fix the no return now

PR #279, opened 2018-06-05
Fixed pagination kwarg bug

mitchmcdee:
Cursor was never being updated because we were updating the wrong kwarg. Also patched an unchecked get for `members`.


TRManderson:
Good to go 👍

PR #283, opened 2018-06-06
Reconnect if connection is dropped

TRManderson:
There was an error recently when the connection was dropped - this attempts a reconnection


mitchmcdee:
Code has been tested?
Also, what happens if we have an exception during `connect`? Should we attempt to recover from that?

TRManderson:
Error gets reported, no recovery is made, supervisorctl restarts process

TRManderson:
Code has been tested but no new code paths have

TRManderson:
Actually this is probably unnecessary now that supervisorctl is fixed

Issue #282, opened 2018-06-06
Enforce async?


mitchmcdee:
Could we save a lot of code complexity if we just enforce all commands to be async functions? Are there any downsides to this?

TRManderson:
Async scripts have a decent chunk of complexity in them. Not sure it's the most beginner friendly.

mitchmcdee:
We don't have to enforce async implementation perse, just use `async def` is more what I mean

TRManderson:
If you use `async def` but don't have an async impl, it clogs up the whole bot waiting for your requests to finish.

mitchmcdee:
wat

TRManderson:
See #285 for what's required for asyncio to actually be a benefit rather than a hindrance

mitchmcdee:
Moving this comment over from #285:

```[I'm] a bit concerned about the complexity and readability sacrifices we're making in exchange for async; ideally that's why I was asking whether it was feasible to enforce async (#282) for everything so that we can abstract things such as with as_async as well as simplify the core code base```

No need to respond, just having it documented. just something we should think about/discuss at some point :^)

PR #285, opened 2018-06-06
Utilise asyncio more

TRManderson:
Most async scripts were using `bot.post_message` synchronously, which would block the thread running tasks.

This PR introduces `bot.as_async`, which is a fixed version of `AsyncBotWrapper` that runs all functions using the event loop's threadpool executor, and provides an async `api` property.

### DIFF
--- a/uqcsbot/base.py
+++ b/uqcsbot/base.py
@@ -139,8 +139,8 @@ class UQCSBot(object):
         return APIWrapper(self.client)
 
     @property
-    def async(self):
-        return AsyncBotWrapper(self.client)
+    def as_async(self):
+        return AsyncBotWrapper(self)
 
     def post_message(self, channel: Union[Channel, str], text: str, **kwargs):
         channel_id = channel if isinstance(channel, str) else channel.id
@@ -220,6 +220,9 @@ class UQCSBot(object):
         self._client = SlackClient(api_token)
         self._verification_token = verification_token
         with self._async_context():
+            # Initialise channels at start so we don't have to block
+            self.channels._initialise() 
+
             if not self.client.rtm_connect():
                 raise OSError("Error connecting to RTM API")
             while True:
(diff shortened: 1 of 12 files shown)


mitchmcdee:
Without checking, how does this differ to `bot.run_async`?

TRManderson:
The API just looks way more natural basically, go check out the changes in one of the scripts that used run async

mitchmcdee:
do we need `run_async` anymore then? can it be removed?

TRManderson:
Still need it for requests.get

uqcsbot/base.py
```diff
@@ -260,9 +263,17 @@ def __init__(self, bot: UQCSBot):
 
     def __getattr__(self, name):
         attr = getattr(self.bot, name)
-        if not callable(attr) or not asyncio.iscoroutinefunction(attr):
+        if not callable(attr) or asyncio.iscoroutinefunction(attr):
```
mitchmcdee: what was the reason for removing the `not` here?

uqcsbot/base.py
```diff
@@ -139,8 +139,8 @@ def api(self):
         return APIWrapper(self.client)
 
     @property
-    def async(self):
-        return AsyncBotWrapper(self.client)
+    def as_async(self):
+        return AsyncBotWrapper(self)
```
mitchmcdee: why the swap from client to self?

TRManderson: It's a bot wrapper, not a client wrapper. Having `self.client` there was always a bug.

uqcsbot/base.py
TRManderson: Fixing the behaviour - if it's a coroutine function we want to return it straight up because it's already async

TRManderson:
Nope, user has to use `as_async`

mitchmcdee:
A bit concerned about the complexity and readability sacrifices we're making in exchange for async; ideally that's why I was asking whether it was feasible to enforce async (#282) for everything so that we can abstract things such as with `as_async` as well as simplify the core code base

I'll merge, but I think its something we should think about and discuss more moving forward

PR #287, opened 2018-06-07
Added !whatsdue

### LINKED ISSUE #158: !calendar breaks on date format 'DD/MM'
edge case for classes like STAT2203, should be a simple fix-- will do it when I find the time (if no one else already has)

### LINKED ISSUE #252: Fix hardcoded exam dates in !calendar
https://github.com/UQComputingSociety/uqcsbot/blob/8a19b88b7d2d2250764e2c394c1270d28bfe2304/scripts/whatsdue.js#L127

mitchmcdee:
Here's the deets:
* Ported !whatdue over
* New and improved
* Automatic exam period detection
* Automatically culls assignments that have already passed
* Added -f/--full to allow for full assignment list printouts
* Made a sieck uq course util library that will make adding `!calendar` super easy.
* Swapped some stuff out in `!ecp` to use the util library instead.

Demo GIFs:
1) Course channel (with --f)
image

2) Specified course (with --full)
image

3) Multiple courses
image

4) Too many courses
image

Fixes #158. Fixes #252.


mitchmcdee:
It should also be mentioned that the utility library I've written here makes it super easy to add `!calendar` in a follow up PR

uqcsbot/scripts/uq_course_util.py
```diff
@@ -0,0 +1,176 @@
...
+BASE_CALENDAR_URL = 'http://www.uq.edu.au/events/calendar_view.php?category_id=16&year='
+
+class DateSyntaxException(Exception):
+    '''
+    Raised when an unparsable date syntax is encountered.
+    '''
+    def __init__(self, date):
+        super().__init__()
+        self.date = date
+        bot.logger.error(f'Could not parse date \'{date}\'.')
```
TRManderson: I really don't like the `bot.logger.error` calls in the constructor - you can replace them with setting the message in the superclass call

TRManderson: An example of a way I'd prefer it:
```python
def __init__(self, date):
    self.message = f'Could not parse date \'{date}\'.'
    self.date = date
    super().__init__(self.message, self.date)
```
and then maybe also
```python
def __str__(self):
    return self.message
```

uqcsbot/scripts/whatsdue.py
```diff
@@ -0,0 +1,68 @@
...
+    # If no full output specified, set the cutoff to today's date. Else, set the
+    # cutoff to UNIX epoch (i.e. filter nothing).
+    cutoff_datetime = datetime.fromtimestamp(0) if is_full_output else datetime.today()
+    try:
+        assessment = get_course_assessment(course_names, cutoff_datetime)
+    except HttpException as e:
+        bot.post_message(channel, f'An error occurred, please try again.')
+        return
+    except CourseNotFoundException as e:
+        bot.post_message(channel, f'Could not find course `{e.course_name}`.')
```
TRManderson: If we make my suggested change from earlier we can turn this into `bot.post_message(channel, e.message)` (or `str(e)` if you decide to do that - I'm not so sure on that one being the right call). Can then just reraise for logging purposes.

PR #299, opened 2018-06-09
Removed async as a core part of the framework

### LINKED ISSUE #282: Enforce async?

### LINKED ISSUE #292: Remove async
After discussion on Slack, as well as in #282 and #250, the decision has been made to remove async from the bot for two main reasons:

1) Reduce code base complexity (and subsequently improve code base readability). Supporting both async and sync lead to a lot of hacky and magic implementations within the core of pybot and made it unnecessarily complex and difficult to understand. This goes against the entire point of our shift to pybot from jsbot, where we wanted the entire codebase (making scripts and the core) to be as beginner friendly as possible. Furthermore, async required additional user understanding to execute effectively within a script (i.e. remembering to use `run_async`, `as_async`, `async def`, `await`), which further decreased usability and readability.

2) We don't benefit from its performance gains. Async is useful when handling many commands, which uqcsbot rarely has to do.

As a result of removing async, we will have significantly trimmed down the code base, significantly reduced complexity, made it easier to make scripts and all for a negligible performance hit.

@TRManderson has kindly volunteered to champion the removal :)

TRManderson:
Swap all the core stuff for a single function `bot.get_event_loop()`, which allows you to use asyncio within the scope of a single command by getting an event loop.

All commands are run in threads using a `concurrent.futures.ThreadPoolExecutor`.

Coroutines as commands no longer works.

Fixes #282. Fixes #292.

### DIFF
--- a/uqcsbot/api.py
+++ b/uqcsbot/api.py
@@ -49,10 +49,6 @@ class APIMethodProxy(object):
         Perform the relevant API request. Equivalent to SlackClient.api_call
         except the `method` argument is filled in.
 
-        If the APIMethodProxy was constructed with `is_async=True` runs the
-        request asynchronously via:
-            asyncio.get_event_loop().run_in_executor(None, req)
-
         Attempts to retry the API call if rate-limited.
         """
         fn = partial(
@@ -75,8 +71,7 @@ class APIMethodProxy(object):
 
     def paginate(self, **kwargs) -> Paginator:
         """
-        Returns a `Paginator` which can be used as both a synchronous and an
-        asynchronous iterable, allowing you to iterate over each page of
+        Returns a `Paginator` which allows you to iterate over each page of
         response data from a Slack response that is paginated in the
         cursor-style.
 
@@ -109,12 +104,6 @@ class APIWrapper(object):
     Example usage:
         > api = APIWrapper(client)
         > api.chat.postMessage(channel="general", text="message")
-
-        > async_api = APIWrapper(client, True)
-        > await async_api.chat.postMessage(channel="general", text="message")
-
-        > async_api = api(is_async=True)
-        > await async_api.chat.postMessage(channel="general", text="message")
     """
     def __init__(self, client: SlackClient) -> None:
         self._client = client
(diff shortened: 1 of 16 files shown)


TRManderson:
Fixes #282 and #292

uqcsbot/api.py
```diff
@@ -186,10 +158,10 @@ def members(self) -> List[str]:
         if self._member_ids is not None:
             return self._member_ids
         self._bot.logger.debug(f"Loading members for {self.name}<{self.id}>")
-        members = []
+        members_ls = []  # type: List[str]
```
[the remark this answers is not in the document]
mitchmcdee: and (more for curiosity) why specify the type here? and isn't `List[str]` incorrect?

uqcsbot/base.py
```diff
@@ -96,18 +91,19 @@ def _handle_goodbye(self, evt):
...
-        ])
+        for handler in self.command_registry[command.command_name]:
+            self.executor.submit(
+                self._execute_catching_error,
+                handler,
+                command,
+            )
 
     def on_command(self, command_name: str):
         def decorator(fn):
```
mitchmcdee: any reason we need `wraps` for `statuc_reacts` but not here?

uqcsbot/base.py
```diff
@@ -138,57 +133,35 @@ def api_call(self, *args, **kwargs):
...
             self._scheduler.shutdown()
             self._executor.shutdown()
-            fix_future.cancel()
-            self._evt_loop.stop()
-            async_thread.join()
-            self._evt_loop.close()
             raise
 
-    async def _await_error(self, awaitable):
+    def _execute_catching_error(self, handler, evt):
```
mitchmcdee: can you provide comments here (and wherever else that is relevant in the scope of this PR) explaining what the function does?

uqcsbot/base.py
```diff
@@ -96,18 +91,19 @@ def _handle_goodbye(self, evt):
             self.logger.debug(f"Goodbye event has unexpected extras: {evt}")
         self.logger.info(f"Server is about to disconnect")
 
-    async def _handle_command(self, message: dict) -> None:
+    def _handle_command(self, message: dict) -> None:
```
mitchmcdee: what's the difference between this `handle_command` and `run_handlers`? docs for both would be nice

uqcsbot/base.py
```diff
@@ -198,10 +171,11 @@ def _run_handlers(self, event):
         if "type" not in event:
```
mitchmcdee: (repeating from above):

what's the difference between this `handle_command` and `run_handlers`? docs for both would  be nice

uqcsbot/api.py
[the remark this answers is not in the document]
TRManderson: The change was because my `mypy` linter in Sublime was complaining, and `List[str]` is correct because it's just member IDs

uqcsbot/base.py
TRManderson: `_handle_command` does the extra execution stuff to get a command obj, whereas `_run_handlers` does it with the raw JSON message

mitchmcdee: cool cool, docs explaining that for both would be great (Y)

PR #302, opened 2018-06-09
Finally merge Pybot to master

TRManderson:
Now that we're only running Pybot we can totally just merge it to master, rather than keeping the `python-rewrite` branch as our main branch.


mitchmcdee:
Conflicts + what backend stuff needs to be setup to make it pull from master again?

TRManderson:
conflicts are easy: just need to do manual merge rather than clicking the button, delete all conflicting files

backend stuff:
delete uqcsbot and uqcsbot-deploy jobs, change the branch on uqcsbot2 to be master, rename job to uqcsbot. delete `/srv/supervisord-hubot.conf` on server.

mitchmcdee:
kk will do when I don't have 3202 study to do <3

PR #303, opened 2018-06-09
Fixed a few things

mitchmcdee:
This PR does a few lil things:
* Added logging of API call errors
* Fixed misnamed repr variable
* Cleaned up some PEP 8 issues
* Removed unused `update_members`
* Patched lingering bad merge issues (`Command.from_message`)

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -44,7 +44,7 @@ class MockUQCSBot(UQCSBot):
         self.channels = {}  # Allows get to fail. TODO mock channel object
 
     def test_handle_event(self, message):
-        command = Command.from_message(self, message)
+        command = Command.from_message(message)
         if command is None:
             raise UnparsedCommandException()
         handlers = self._command_registry[command.command_name]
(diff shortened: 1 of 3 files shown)


TRManderson:
Honestly I reckon we should just implement a user data store that lazily loads this kind of thing, rather than doing a query for all N users on startup (re deleted)

mitchmcdee:
I thought it was already lazily loaded tbh
also fwiw its async

mitchmcdee:
@TRManderson we already have the channel wrapper which is our "user data store" (I don't think a dedicated user data store is necessary) and that lazily loads members when requested. This PR expands that lazy loading to also check for active members

TRManderson:
The list of members for a channel is lazily loaded. When this occurs, it triggers the eager (but async) loading of all member info for that channel, just to check the `is_active` property. On #announcements this means 1200 requests

I am suggesting we lazily load the members for a channel, and when we need member activity info, lazily load that info into a separate data store.

mitchmcdee:
I feel ya. sgtm. will make the changes soooooon

TRManderson:
Also should correspond to a change to wakie that samples, _then_ checks whether or not a member is active.

Unfortunately milestone counts will still trigger a metric fuckload of requests.

TRManderson:
Also if you don't get around to it by tomorrow I'm totally going to have a crack after my 8am exam

TRManderson:

mitchmcdee:
go for it :^) I'll remove it from this PR

mitchmcdee:
can ya make an issue for it?

PR #304, opened 2018-06-09
WIP: Added !supervisor for supervisord process control and admin_only decorator

### LINKED ISSUE #88: !logs {procname}
All UQCS stuff is managed by supervisor on our server. Supervisor provides an XML-RPC endpoint that we can use to manage the processes on the server, and is used by cesi.

Using the XML-RPC endpoint, we should be able to view the tail of the logs of a given process. See hookbot (https://github.com/UQComputingSociety/webhooks) for an example of using the XML-RPC endpoint (in Python).

Supervisor docs are here: http://supervisord.org/api.html

mitchmcdee:
This PR accomplishes the long standing issue #88 and improves on the idea by supporting all core supervisor functionality. Through this script a user is able to start, stop and restart a supervisord process as well as get its state and log file tail.

Obviously, this is a script that could be easily abused in the wrong hands. As such, I've also included in this PR the `admin_only` feature, which is a decorator that can be added to scripts to make them only executable by admins.

Fixes #88.

admin only:
image

tail:
image

stop and info:
image

### DIFF
--- a/uqcsbot/scripts/supervisor.py
+++ b/uqcsbot/scripts/supervisor.py
@@ -1,0 +1,72 @@
+import xmlrpc.client
+import argparse
+from uqcsbot import bot, Command
+from uqcsbot.utils.command_utils import (loading_status, success_status,
+                                         FailureException, admin_only,
+                                         UsageSyntaxException)
+
+
+def handle_supervisor_command(parsed_args, channel_id) -> None:
+    '''
+    Processes and executes the supervisor command.
+    '''
+    server = xmlrpc.client.Server('http://localhost:9001/RPC2')
+    command_name = parsed_args.command_name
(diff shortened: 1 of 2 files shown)


mitchmcdee:
Not so sure about my latest commit-- it cleaned up the code a lot from the previous state, but idk about the whole failure reacts over just posting to the channel that something fucked up. honestly just pretty tired atm so bedtime for me and i'll look at it later.

mitchmcdee:
Also it should be noted that this doesn't make it possible to start uqcsbot when it has crashed and needs manual restarting 😂

TRManderson:

mitchmcdee:
Retargeted to base branch

mitchmcdee:
Will fix this up once #334 lands as that will be useful here

PR #310, opened 2018-06-10
Added exception handling for unparsable dates

mitchmcdee:
This PR makes it so that if we can't parse a due date, we assume that it has not passed yet and return it to the user. This is better than the alternative of just not showing it at all.

Ideally, we want to be able to parse all potential due dates. As such, I've added a TODO for this; but this PR fix should suffice for now.


TRManderson:

mitchmcdee:
Retargeted to base branch

PR #312, opened 2018-06-12
Add !hoogle

MaxwellBo:
*With newlines for `--verbose`*
withnewlines

*Without newlines for `--verbose`*
withoutnewlines

### DIFF
--- a/uqcsbot/scripts/hoogle.py
+++ b/uqcsbot/scripts/hoogle.py
@@ -1,0 +1,57 @@
+from uqcsbot import bot, Command
+import requests
+import json
+import html
+
+def get_endpoint(type_sig: str) -> str:
+    unescaped = html.unescape(type_sig)
+
+    return "https://www.haskell.org/hoogle/?mode=json&hoogle=" + unescaped + "&start=0&count=10"
+
+def pretty_hoogle_result(result: dict, is_verbose: bool) -> str:
+    url = result['location']
+    type_sig = result['self']
+    docs = result['docs']
+
+    if is_verbose:
+        return f"`{type_sig}` <{url}|link> {docs}"
+    else:
+        return f"`{type_sig}` <{url}|link>"
+        
+@bot.on_command("hoogle")
+def handle_hoogle(command: Command):
+    '''
+    `!hoogle [--verbose] <TYPE_SIGNATURE>` - Queries the Hoogle Haskell API search engine, 
+    searching Haskell libraries by either function name, or by approximate type signature.
+    '''
+    command_args = command.arg.split() if command.has_arg() else []
+
+    verbose = False
+
+    if '--verbose' in command_args:
+        command_args.remove('--verbose')
+        verbose = True
+
+    if len(command_args) == 0:
+        bot.post_message(command.channel, handle_hoogle.__doc__)
+        return
+
+    type_sig = ' '.join(command_args)
+
+    endpoint_url = get_endpoint(type_sig)
+    
+    http_response = requests.get(endpoint_url)
+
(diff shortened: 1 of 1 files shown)


uqcsbot/scripts/hoogle.py
```diff
@@ -0,0 +1,57 @@
...
+
+    return "https://www.haskell.org/hoogle/?mode=json&hoogle=" + unescaped + "&start=0&count=10"
+
+def pretty_hoogle_result(result: dict, is_verbose: bool) -> str:
+    url = result['location']
+    type_sig = result['self']
+    docs = result['docs']
+
+    if is_verbose:
+        return f"`{type_sig}` <{url}|link> {docs}"
```
TRManderson: Formatting  could definitely be better for verbose mode

MaxwellBo: `\n` between the link and the docs?

uqcsbot/scripts/hoogle.py
```diff
@@ -0,0 +1,57 @@
...
+    command_args = command.arg.split() if command.has_arg() else []
+
+    verbose = False
+
+    if '--verbose' in command_args:
+        command_args.remove('--verbose')
+        verbose = True
+
+    if len(command_args) == 0:
+        bot.post_message(command.channel, handle_hoogle.__doc__)
```
mitchmcdee: I like this, good use of `__doc__` 👍 i reckon we'll adopt something similar for other scripts
maybe prepend with `usage: ` just to be 100% clear?

uqcsbot/scripts/hoogle.py
mitchmcdee: throw up a screenshot of both options?

mitchmcdee: (so we can see + help decide)

uqcsbot/scripts/hoogle.py
```diff
@@ -0,0 +1,57 @@
...
+def handle_hoogle(command: Command):
+    '''
+    `!hoogle [--verbose] <TYPE_SIGNATURE>` - Queries the Hoogle Haskell API search engine, 
+    searching Haskell libraries by either function name, or by approximate type signature.
+    '''
+    command_args = command.arg.split() if command.has_arg() else []
+
+    verbose = False
+
+    if '--verbose' in command_args:
```
mitchmcdee: might as well add `-v` too

TRManderson: may as well just use argparse at this point 😂

mitchmcdee: yis.

MaxwellBo: isn't `-v` always for version?

TRManderson: curl, tar, mv, cp

uqcsbot/scripts/hoogle.py
MaxwellBo: Added screenshots

uqcsbot/scripts/hoogle.py
mitchmcdee: I thought `-V` was version i.e. python

uqcsbot/scripts/hoogle.py
mitchmcdee: `\n` lgtm

uqcsbot/scripts/hoogle.py
MaxwellBo: ah fair enough, I'll add `-v` then

PR #315, opened 2018-06-13
Added auto reconnect and no team state flags

mitchmcdee:
The slack client library we're using provides auto-reconnect logic for us that we weren't using. This PR adds that logic through setting the `auto_reconnect` flag and also sets the `with_team_state` flag to `False` for a lighter connection (which is recommended for larger teams such as ours)

See: https://slackapi.github.io/python-slackclient/real_time_messaging.html#rtm-start-vs-rtm-connect

### DIFF
--- a/uqcsbot/base.py
+++ b/uqcsbot/base.py
@@ -213,7 +213,7 @@ class UQCSBot(object):
             # Initialise channels at start so we don't have to block
             self.channels._initialise()
 
-            if not self.client.rtm_connect():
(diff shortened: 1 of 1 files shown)


mitchmcdee:
This was fixed in Mar 31 ish as can be seen here https://github.com/slackapi/python-slackclient/issues/118

TRManderson:
Didn't realise there was good shit in `rtm.start`! Doing data stores is probably a better long-term approach though. LGTM 👍

PR #319, opened 2018-06-15
User data store

### LINKED ISSUE #190: Scripts that use the conversations.members API method don't check if a user is deleted/deactivated
as per title, would need to do an extra API call to check user status :(

### LINKED ISSUE #314: Add lazy loading of user information
Removed from #303 (look there for discussion), assigning to @TRManderson

TRManderson:
First pass at a user data store - did some very rudimentary testing but it's not enough that I'm super duper confident about it.

Fixes #314, kind of. Doesn't lazy-load the data, but uses the RTM API connection package to initialise, and RTM events to track state changes. Also updates the channel wrapper to do the same.

There's a fair bit of duplicated logic between the wrappers that I'd eventually like to fix, but getting this to land I feel is marginally more important.

Fixes #190.


uqcsbot/api.py
```diff
@@ -224,6 +232,19 @@ def _initialise(self):
                     self._add_channel(im)
             self._initialised = True
 
+    def populate_from_team_state(self, data):
+        with self._lock:
+            self._initialised = True
+            self._channels_by_id = {}
+            self._channels_by_name = {}
+            chan_types = ['channels', 'groups', 'ims']
```
mitchmcdee: just inline this

uqcsbot/api.py
```diff
@@ -305,5 +326,130 @@ def _on_group_joined(self, evt):
     def _on_group_left(self, evt):
         self._on_channel_deleted(evt)
 
-class MembersWrapper(object):
-    pass
+
+class UsersWrapper(object):
```
mitchmcdee: if changing to `users` from `members` (which I agree with) can you change it everywhere its mentioned in the repo pls

uqcsbot/api.py
```diff
@@ -305,5 +326,130 @@ def _on_group_joined(self, evt):
...
+            self._bot.on(mtype, attr)
+
+    def _on_user_change(self, evt):
+        user = self._users_by_id[evt['user']['id']]
+        old_display_name = user.display_name
+        with self._lock:
+            user.update_from_dict(evt['user'])
+            if user.display_name != old_display_name:
+                self._users_by_display[old_display_name].remove(user)
+                if len(self._users_by_display[old_display_name]) == 0:
```
mitchmcdee: can you add a comment here explaining code intent because its not immediately obvious

uqcsbot/api.py
mitchmcdee: i.e. `member_ids`

mitchmcdee:
Just saw your comment re: testing lel-- do you want to do anything extra before merging? just more manual tests?

uqcsbot/api.py
TRManderson: Gonna maybe make a base class for this shit seeing it's basically the same on users and channels

TRManderson:
It's all good re testing, just needs some more testing before it gets any real use. Milestones is also missing the fix

mitchmcdee:
Another thing: could you add logging for the number of members when we calculate it for milestone so we can see how many users it thinks we're at (in case the big 1600 fails)

uqcsbot/api.py
TRManderson: I kind of feel like the channel one should stay as is tbh - it's talking about the members of a generic "conversation", whereas the user data store stores info about all users

uqcsbot/api.py
TRManderson: OK I'm not gonna make a base class for this

uqcsbot/api.py
mitchmcdee: sgtm

uqcsbot/api.py
```diff
@@ -305,5 +325,131 @@ def _on_group_joined(self, evt):
...
+        self._initialised = False
+        self._initialise()
+
+    def get(self, display_name_or_id: str, default: T = None, use_cache: bool =True) -> Union['User', List['User'], T]:
+        if use_cache and not self._initialised:
+            self._initialise()
+        if display_name_or_id in self._users_by_id:
+            return self._users_by_id[display_name_or_id]
+        elif display_name_or_id in self._users_by_display:
+            return self._users_by_display[display_name_or_id]
```
mitchmcdee: I just realised, why are we storing users by their display name? like, won't this `get` return a list of users?

TRManderson: yup, this will return a list of users (the type reflects that)

Unfortunately display names can collide, but that's there in case we end up needing them for something

mitchmcdee: would prefer to just remove them tbh, unless you have an example where you think its necessary? lets just use user ids

uqcsbot/scripts/welcome.py
```diff
@@ -31,17 +31,26 @@ def welcome(evt: dict):
...
+    if user and not user.is_bot:
+        bot.executor.submit(welcome_messages)
+
+    if user:
+        bot.post_message(general, f"Welcome, {user.display_name}")
+
+    valid_users = len([
+        member_id
+        for member_id in announcements.members
+        if not getattr(bot.users.get(member_id), "deleted", True)
```
mitchmcdee: why not just `if not bot.users.get(member_id)['deleted']`

mitchmcdee: (the getattr stuff is neat but a bit of a confusing mouthful)

uqcsbot/scripts/welcome.py
```diff
@@ -31,17 +31,26 @@ def welcome(evt: dict):
...
+    ])
+    bot.logger.info(f"Currently at {valid_users} members")
+    if valid_users % MEMBER_MILESTONE == 0:
         bot.post_message(
```
mitchmcdee: can probs make this one line now

mitchmcdee: i.e. `bot.post_message(general, f":tada: {valid_users} members! :tada:")`

uqcsbot/scripts/welcome.py
```diff
@@ -31,17 +31,26 @@ def welcome(evt: dict):
...
-        bot.post_message(evt.get("user"), message)
-    if len(announcements.members) % MEMBER_MILESTONE == 0:
+    user = bot.users.get(evt.get("user"))
+
+    def welcome_messages():
+        for message in WELCOME_MESSAGES:
+            time.sleep(MESSAGE_PAUSE)
+            bot.post_message(evt.get("user"), message)
+    if user and not user.is_bot:
+        bot.executor.submit(welcome_messages)
```
mitchmcdee: `executor` in scripts makes me sad, is it super necessary here?

uqcsbot/scripts/welcome.py
mitchmcdee: sgtm, add a comment with that explanation pls :^)

mitchmcdee:
Fix conflicts and its g to merge

PR #324, opened 2018-06-16
Implement !latex

TRManderson:


test/test_latex.py
```diff
@@ -0,0 +1,12 @@
+"""
+Tests for uqcsbot.scripts.cat
+"""
+from .conftest import MockUQCSBot
+from .helpers import generate_message_object
+
+
+def test_latex(uqcsbot: MockUQCSBot):
```
mitchmcdee: this test needs updating to the new test framework (also, can you adopt the new styling too)

uqcsbot/scripts/latex.py
```diff
@@ -0,0 +1,42 @@
...
+                "id": 1,
+            }
+        ],
+        unfurl_links=False,
+        unfurl_media=False,
+    )
+
+
+@bot.on_command("latex")
+def handle_latex_cmd(cmd: Command):
```
mitchmcdee: nit: prefer `command` over `cmd` to align with other scripts

uqcsbot/scripts/latex.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    `!latex CONTENT` - Renders `CONTENT` to LaTeX and sends it to Slack.
+    `$$ CONTENT $$` also works.
+    """
+    if not cmd.has_arg():
+        bot.post_message(cmd.channel, "No data provided")
```
mitchmcdee: needs to be updated to `cmd.channel_id` here and below

uqcsbot/scripts/latex.py
```diff
@@ -0,0 +1,42 @@
...
+    url = f"http://latex.codecogs.com/gif.latex?{quote(data)}"
+    bot.post_message(
+        channel, 
+        text=f"LaTeX render for \"{data}\"",
+        attachments=[
+            {
+                "fallback": f"Codecogs LaTeX image for {data}",
+                "image_url": url,
+                "from_url": url,
+                "id": 1,
```
mitchmcdee: what's `id` used for? (is it required?)

TRManderson: No idea, I think it's just for ordering? I copied it from a deconstructed `message_changed` event.

mitchmcdee: strip out as much as possible imo, if it ain't needed turf it

test/conftest.py
```diff
@@ -166,15 +166,15 @@ def mocked_chat_postMessage(self, **kwargs):
...
         channel = self.channels.get(channel_id_or_name)
         if channel is None:
             return {'ok': False}
 
-        message = {'text': text, 'ts': str(time.time())}
+        message = {'ts': str(time.time()), **kwargs}
         self.test_messages[channel.id].append(message)
-        message_event = {'type': 'message', 'channel': channel.id, **message}
+        message_event = {'type': 'message', **message}
+        message_event['channel'] = channel.id
```
mitchmcdee: why is this brought outside?

test/test_latex.py
```diff
@@ -1,12 +1,13 @@
 """
 Tests for uqcsbot.scripts.cat
 """
```
mitchmcdee: can you move this to be a function doc instead like the other tests? (and update it from being for !cat)

test/test_latex.py
```diff
@@ -1,12 +1,13 @@
...
 def test_latex(uqcsbot: MockUQCSBot):
-    message = generate_message_object("!latex test_memes")
-    uqcsbot.test_handle_event(message)
-    assert len(uqcsbot.test_posted_messages) == 1
-    assert uqcsbot.test_posted_messages[-1].text == "LaTeX render for \"test_memes\""
\ No newline at end of file
+    uqcsbot.post_message(TEST_CHANNEL_ID, "!latex test_memes")
+    messages = uqcsbot.test_messages.get(TEST_CHANNEL_ID, [])
+    assert len(messages) == 2
+    print(messages)
```
mitchmcdee: remove print

test/test_latex.py
```diff
@@ -1,12 +1,13 @@
...
+    assert messages[-1]["text"] == "LaTeX render for \"test_memes\""
+    assert len(messages[-1].get("attachments", [])) == 1
```
mitchmcdee: new line

mitchmcdee: also why not something like `messages[-1].get('attachments') is not None`?

test/conftest.py
TRManderson: The message event is what the bot would receive, which I'm pretty sure is always a channel id. The splat was overwriting it with whatever the arg passed in was.

test/test_latex.py
TRManderson: I want it to return exactly 1 attachment

mitchmcdee:
Tested and ready to merge?

TRManderson:
Yep! `pytest` and real-world tests.

PR #332, opened 2018-06-17
Fixed all lint issues. Added coverage report output. Added test linting.

### LINKED ISSUE #291: !wolfram should use the env variable and UQCS app id
The app id for !wolfram is currently hard coded into the script and is also my personal app id. I don't really mind but I assume it would be preferable to have a specific UQCS one?

I believe the environment variable already exists as "WOLFRAM_APPID".

Happy to do this after exams.

mitchmcdee:
As per title :^)

Fixes #291

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -4,6 +4,7 @@ Configuration for Pytest
 
 from unittest.mock import MagicMock
 from itertools import islice
+from functools import partial
 from collections import defaultdict
 import time
 import pytest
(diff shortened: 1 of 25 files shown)


mitchmcdee:
Note: this PR now adds strict line limits + import usage since I cleaned up all existing issues.

TRManderson:
It seems to me like the verdict here is to prefer `# noqa` for unused imports from type comments - are we locking that in?

mitchmcdee:
Means we can catch legitimate unused imports so yeah would prefer `# noqa` imo

test/conftest.py
```diff
@@ -132,17 +132,23 @@ def mocked_channels_list(self, channel_type='channels', **kwargs):
         cursor = kwargs.get('cursor', 0)
         limit = kwargs.get('limit', 100)
 
+        def is_channel_type(channel, channel_type):
```
TRManderson: `is_channel_type = partial(dict.get, channel, default=False)`

mitchmcdee: ye nice

TRManderson: Nah doesn't work

TRManderson:
```python
>>> from functools import partial
>>> x = {}
>>> fn = partial(dict.get, x, default=False)
>>> fn(43)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: get() takes no keyword arguments
>>> fn(43)
```

mitchmcdee: kek. good to merge then?

PR #333, opened 2018-06-17
Fix test_yt

jdcaperon:
This PR fixed the yt.py tests by mocking the google-api-client. It also abstracts out the execution of search on the youtube API to allow for better testing and future functionality.

Super basic, but it makes things green for now.


test/test_yt.py
```diff
@@ -11,14 +13,33 @@
...
 
+def mocked_search_execute(search_query: str, search_part: str, search_type: str, max_results: int):
+    """
+    Currently only returns a response of video ID's based on max_results.
+    Otherwise returns none.
+    """
+    if search_type == 'video' and search_part == 'id':
+        items = []
+        for _ in range(max_results):
+            # The following line generates an 11 character random string of ascii and digits
```
mitchmcdee: since we don't actually use the ids for anything, you should probably just return a zero padded increasing int or some shit
i.e. `items = [{'id': {'videoId': str(i).zfill(11)}} for i in range(max_results)`
this just makes the code simpler

test/test_yt.py
```diff
@@ -11,14 +13,33 @@
 NO_QUERY_MESSAGE = "You can't look for nothing. !yt <QUERY>"
 
 
+def mocked_search_execute(search_query: str, search_part: str, search_type: str, max_results: int):
+    """
+    Currently only returns a response of video ID's based on max_results.
+    Otherwise returns none.
+    """
+    if search_type == 'video' and search_part == 'id':
```
mitchmcdee: flip this if statement to handle the negative case first so we can remove a layer of indentation :^)

test/test_yt.py
```diff
@@ -11,14 +13,33 @@
...
+            videoId = ''.join(random.choices(
+                string.ascii_letters + string.digits, k=11))
+            # The response from the client contains a list of items
+            # Each item has id object containing a string called videoId
+            items.append({'id': {'videoId': videoId}})
+        return {'items': items}
+    return None
+
+
 def test_yt_no_query(uqcsbot: MockUQCSBot):
```
mitchmcdee: explain test intent here

test/test_yt.py
```diff
@@ -11,14 +13,33 @@
...
     uqcsbot.post_message(TEST_CHANNEL_ID, "!yt")
     messages = uqcsbot.test_messages.get(TEST_CHANNEL_ID, [])
     assert len(messages) == 2
     assert messages[-1]['text'] == NO_QUERY_MESSAGE
 
 
-@pytest.mark.skip
+@patch("uqcsbot.scripts.yt.execute_search", new=mocked_search_execute)
 def test_yt_normal(uqcsbot: MockUQCSBot):
```
mitchmcdee: explain test intent here

test/test_yt.py
jdcaperon: I'm not sure I understand

mitchmcdee:
```
if search_type != 'video' or search_part != 'id':
    return None
return {'items': [{'id': {'videoId': str(i).zfill(11)}} for i in range(max_results)]}
```

mitchmcdee: made more sense before new changes-- but I'm always a fan of handling the negative case first (i.e. invalid cases)

test/test_yt.py.orig
```diff
@@ -0,0 +1,56 @@
+"""
```
mitchmcdee: delete this file ;)

uqcsbot/scripts/yt.py
```diff
@@ -36,16 +36,16 @@ def handle_yt(command: Command):
         bot.post_message(command.channel_id, "Your query returned no results.")
 
 
-def get_top_video_result(search_query: str):
+def get_top_video_result(search_query: str, channel):
     '''
     The normal method for using !yt searches based on query
     and returns the first video result. "I'm feeling lucky"
     '''
     search_response = execute_search(search_query,
```
mitchmcdee: put the comment above
i.e. "Return one video id."

uqcsbot/scripts/yt.py.orig
```diff
@@ -0,0 +1,82 @@
+import os
```
mitchmcdee: delete this file :^)

uqcsbot/scripts/yt.py
mitchmcdee: (the inline comments make things a bit untidy, and I reckon the explanations are simple enough to chuck it above)

test/test_yt.py
mitchmcdee: will wait on your response on this before merging :^) @jdcaperon

mitchmcdee:
@jdcaperon still waiting on your input in https://github.com/UQComputingSociety/uqcsbot/pull/333/files#diff-456fb7eb5006535fbe3432ab1f0833a2 before merge FYI :^)

test/test_yt.py
jdcaperon: Understood, yeah in the context of the function itself IMO it makes more sense to keep it in its current form. On its own, it looks rather awkward but given the context, I would rather keep it in its current format.

jdcaperon:
Back from China now ;D lgtm

PR #347, opened 2018-07-27
Added !meme

mubiquity:
Generates a meme based on the input. For instance !meme ants "Top" "Bottom" produces:
meme

Also, running !meme names will send a list of all of the allowable names to the users channel. In future UQCS specific memes could be added.

I don't know why I did this.


uqcsbot/scripts/meme.py
```diff
@@ -0,0 +1,308 @@
+from uqcsbot import bot, Command
+import re
+from uqcsbot.util.status_reacts import loading_status, success_status
+
+API_URL = "https://memegen.link/"
+# Many different characters need to be replaced in order to work in url format
+# See the API_URL for details
+REPLACEMENTS = {'_': '__', ' ': '_', r'\"': "''", '-': '--', '?': '~q', '%': '~p', '#': '~h', '/': '~s'}
+VALID_NAMES = [
```
CameronAavik: Could we store the names list and meme names list as a dictionary instead so we don't have the duplicated names? Not really a blocker for merging for me though

uqcsbot/scripts/meme.py
```diff
@@ -0,0 +1,308 @@
...
+        send_meme_names(command)
+        return
+    elif len(args) == 2:
+        top, bottom = args
+    else:
+        bot.post_message(channel, "You supplied the wrong number of args. Please run !help meme")
+        return
+
+    # Make an attachment linking to image
+    image_url = API_URL + f"{name}/{top}/{bottom}.jpg"
```
CameronAavik: Actually, can we make sure to escape the name, top, and bottom args here so we don't get any tricky injection going on

mubiquity: Sorry, not quite sure what you mean. What kind of injection could happen here?

CameronAavik: for example what if the variable "top" contained a forward slash? you need to escape those special characters. pretty sure you want to use `urllib.quote`

Here is an example:
https://github.com/UQComputingSociety/uqcsbot/blob/6a4db88fbf9916984ba3ceb82092f65f58bee74d/uqcsbot/scripts/acronym.py#L14

uqcsbot/scripts/meme.py
```diff
@@ -0,0 +1,308 @@
+from uqcsbot import bot, Command
+import re
+from uqcsbot.util.status_reacts import loading_status, success_status
+
+API_URL = "https://memegen.link/"
+# Many different characters need to be replaced in order to work in url format
+# See the API_URL for details
+REPLACEMENTS = {'_': '__', ' ': '_', r'\"': "''", '-': '--', '?': '~q', '%': '~p', '#': '~h', '/': '~s'}
```
MaxwellBo: Do we want something to do with https://docs.python.org/3/library/urllib.parse.html#url-quoting here?

uqcsbot/scripts/meme.py
mubiquity: Okay I can do that. Though any forward slashes should be replaced with a ~s in the get_meme_arguments function

uqcsbot/scripts/meme.py
mubiquity: Does that allow custom changes? For instance the api expects forward slashes to be encoded in the url as: ~s

CameronAavik: I think using a custom solution is fine tbh, plus it's all explained at memegen.link

test/test_meme.py
```diff
@@ -0,0 +1,11 @@
...
+
+def test_meme(uqcsbot: MockUQCSBot):
+    '''
+    Test !meme
+    '''
+    uqcsbot.post_message(TEST_CHANNEL_ID, '!meme bender "top" "bottom"')
+    messages = uqcsbot.test_messages.get(TEST_CHANNEL_ID, [])
+
+    assert len(messages) == 2
+    assert messages[1]['attachments'][0]['image_url'] == 'https://memegen.link/bender/top/bottom.jpg'
```
mitchmcdee: missing newline

mitchmcdee: prefer `messages[-1]` over `messages[1]` as it implies its the most recent message

uqcsbot/scripts/meme.py
```diff
@@ -0,0 +1,192 @@
...
+
+def get_meme_arguments(input_args: str):
+    """Gets the top and bottom text and returns them in a url friendly form that conforms with the api standards"""
+    # This gets the text between the quotation marks (and ignores \")
+    args = re.findall(r'"(.*?(?<!\\))"', input_args)
+
+    # Replaces all the required characters to be url friendly
+    url_friendly_args = []
+    for arg in args:
+        for old, new in REPLACEMENTS.items():
```
mitchmcdee: https://www.tutorialspoint.com/python/string_translate.htm might be useful here :)

uqcsbot/scripts/meme.py
```diff
@@ -0,0 +1,192 @@
...
+    For a full list of names: !meme names
+    """
+    channel = command.channel_id
+
+    if not command.has_arg():
+        bot.post_message(channel, "Please run !help meme for usage")
+        return
+
+    name = command.arg.split()[0].lower()
+    if name != "names" and name not in MEME_NAMES.keys():
```
mitchmcdee: might as well handle the `if name == names` from below up here

mitchmcdee: especially because it appears it doesn't require any additional checks

uqcsbot/scripts/meme.py
```diff
@@ -0,0 +1,192 @@
...
+        return
+
+    # Make an attachment linking to image
+    image_url = API_URL + f"{quote(name)}/{quote(top)}/{quote(bottom)}.jpg"
+    attachments = [{"text": "", "image_url": image_url}]
+    bot.post_message(channel, "", attachments=attachments)
+
+
+@success_status
+def send_meme_names(command: Command):
```
mitchmcdee: this appears to be the same format as sending an image. Maybe remove it from a function (which looks untidy since you're passing `command` around) and place it back into the if block?

uqcsbot/scripts/meme.py
```diff
@@ -0,0 +1,192 @@
...
+}
+
+# TODO: Would be really simple to add custom UQCS memes
+
+
+@bot.on_command("meme")
+@loading_status
+def handle_meme(command: Command):
+    """
+    !meme <meme name> "<top text>" "<bottom text>"
```
mitchmcdee: would something like using `argparse` and `-t` from top and `-b` for bottom be better maybe? idk up to you. would remove the need for regex which is nice

mitchmcdee:
also, great work :^)

uqcsbot/scripts/meme.py
mubiquity: I took that out so I could add the @success_status unless there is an alternative way or just better to not use it?

uqcsbot/scripts/meme.py
mubiquity: Though about that but didn't want to remove the unlikely case of wanting '-t' to be a part of the result

uqcsbot/scripts/meme.py
mubiquity: Oh yeah, had that at one point but had to remove it for multiple character keys which are no longer needed. Will add back, cheers.

uqcsbot/scripts/meme.py
mitchmcdee: what do you mean so that you can add success status? you can just put it above loading status in `handle_meme` right?

uqcsbot/scripts/meme.py
mitchmcdee: surely argparse can handle that, probably just using quotations like you are now (but at least after you won't need the regex + args handling stuff)

mitchmcdee: argparse isn't that complicated-- might just look a little messy to setup but I think it will improve the code readability a fair bit

if you implement it and decide its too complicated though i'm happy to keep what you currently have :^)

uqcsbot/scripts/meme.py
mitchmcdee: ahhhhhhhh I see-- then yeah this is probably fine, just add a comment explaining this reasoning :^)

uqcsbot/scripts/meme.py
mubiquity: I meant more so when it comes to using it than anything else.

mitchmcdee: yeah also true. ultimately its up to you-- happy to go either way

PR #350, opened 2018-08-08
Remove bots from welcome count

mitchmcdee:
As per title. This move was made such that the member count accurately reflects the number provided by slack-- which seems to ignore bots.

I also moved the time consuming welcome messages to after the milestone calculation as there is a large gap between the "Welcome, {USER}!" and ":tada: XXXX members! :tada:" otherwise.


TRManderson:
Poor UQCSBot :'(

TRManderson:
Caveat is that the member number check has potential to consume a decent chunk of time (though _generally_ less than the welcome DMs)

PR #372, opened 2018-08-23
LOUD WELCOME FOR LOUD USERS

gricey432:
If a user joins whose id is all uppercase e.g. `MITCHELL` then the person will be equally loudly welcomed


gricey432:
lol, fixing tests

gricey432:
I'm also retarded, I'm gonna delete this and do it when I am able to actually test my code

PR #373, opened 2018-08-26
Add Asciify

### LINKED ISSUE #343: Add ASCI text generator
Can use the following API to generate for us
BaseURL = http://artii.herokuapp.com/
Get text "foobar". Request to BaseURL += "make?text=foobar"
Get list of avaliable fonts. Request to BaseURL += "fonts_list"
Get text "foobar" with font "larry3d". Request to BaseURL + "make?text=foobar&font=larry3d"
Using this stuff should make this process pretty simple

adriant4456:
Adds asciify feature, closes #343


mitchmcdee:
apologies for the delayed review, am super busy rn but will try and look at this at some point!

adriant4456:
No problem, I'm still working on improving the tests with unittest mock

uqcsbot/scripts/ascii.py
```diff
@@ -0,0 +1,113 @@
...
+            ascii_text = f"```\n{resp.text}\n```"
+            return ascii_text
+    except RequestException as e:
+        return None
+    
+
+def randomfont() -> str:
+    fontslist = get_fontslist()
+    if fontslist:
+        font_number = random.randint(0, len(fontslist) - 1)
```
TRManderson: `return random.choice(fontslist)` is more concise

uqcsbot/scripts/ascii.py
```diff
@@ -0,0 +1,113 @@
...
+def asciify(text: str, font: str) -> str:
+    try:
+        if font is not None:
+            resp = get(ASCII_URL + text + '&font=' + font)
+            ascii_text = f"```\n{resp.text}\n```"
+            return ascii_text
+        else:
+            resp = get(ASCII_URL + text)
+            ascii_text = f"```\n{resp.text}\n```"
+            return ascii_text
```
TRManderson: the only difference between the branches of the if statement is the URL passed to `get` - move the URL into the `if` on its own and take the rest out

uqcsbot/scripts/ascii.py
```diff
@@ -0,0 +1,113 @@
...
+    except RequestException as e:
+        return None
+    
+
+def randomfont() -> str:
```
TRManderson: you use `randomfont` as a function and `random_font` as a var - maybe `get_random_font` here?

uqcsbot/scripts/ascii.py
```diff
@@ -0,0 +1,113 @@
...
+    fontslist = get_fontslist()
+    if fontslist:
+        font_number = random.randint(0, len(fontslist) - 1)
+        return fontslist[font_number]
+    else:
+        return None
+
+
+def get_fontslist() -> list:
```
TRManderson: honestly you could probably make this a `set` so checking with `in` is faster

uqcsbot/scripts/ascii.py
```diff
@@ -0,0 +1,113 @@
...
+    #check for font list option
+    if '--fontslist' in command_args:
+        return_fonts = True
+        command_args.remove('--fontslist')
+    #check for custom font option
+    fontslist = get_fontslist()
+    if not fontslist:
+        bot.post_message(command.channel_id, ERROR_MESSAGE)
+        return
+    for i in fontslist:
```
TRManderson: there's going to be a lot more fonts than command args so you may as well loop over command args instead of fonts (and break out of the loop once they don't start with `--`)

note that if you do this you need to check for `--randomfont` first

uqcsbot/scripts/ascii.py
```diff
@@ -0,0 +1,113 @@
...
+    if text is None:
+        bot.post_message(command.channel_id, NO_QUERY_MESSAGE)
+        ascii_text = None
+    else:
+        if random_font:
+            ascii_text = asciify(text, randomfont())
+        elif custom_font:
+            ascii_text = asciify(text, selected_font)
+        else:
+            ascii_text = asciify(text, None)
```
TRManderson: I'd probably move the `ascii_text = asciify(text, font)` out of the if statement here and just use the if statement to introduce a `font` variable

test/test_ascii.py
```diff
@@ -0,0 +1,82 @@
...
+"""
+from test.conftest import MockUQCSBot, TEST_CHANNEL_ID
+from unittest.mock import patch
+
+TEST_TEXT = "ThIS iS a TeST MesSAgE"
+FONTSLIST_URL = "http://artii.herokuapp.com/fonts_list"
+
+NO_QUERY_MESSAGE = "Can't ASCIIfy nothing... try `!asciify <TEXT>`"
+BOTH_OPTIONS_MESSAGE = "Font can only be random OR specified"
+ERROR_MESSAGE = "Trouble with HTTP Request, can't ASCIIfy :("
```
TRManderson: Can you just import these reused string contents from the script itself?

uqcsbot/scripts/ascii.py
adriant4456: Thanks, wasn't aware of this function

mitchmcdee:
yeah this is a known issue, just duplicate the strings for now

PR #378, opened 2018-09-06
fixed dm links

nicklambourne:
As per suggestions I've fixed the channel links in the DMs on the #jobs-bulletin watcher.


uqcsbot/scripts/jobs.py
```diff
@@ -21,68 +23,86 @@
...
+    f" exploited, it isn't meant as a personal attack). Additionally, all <{EAIT_UNPAID_JOBS}|unpaid placements> for "
+    f"students in the <{EAIT_FACULTY}|EAIT Faculty> must be approved by the faculty.",
     f"4. Job postings _must_ conform to our <{CODE_OF_CONDUCT}|Code of Conduct> and must not discriminate against "
     f"applicants based on race, religion, sexual orientation, gender identity or age.",
     f"If you have any questions, please get in touch with the committee in #uqcs-meta or by email at "
     f"<{UQCS_EMAIL}|contact@uqcs.org.au>."
 ]
 
 
+def replace_channel_links(messages: List[str]) -> List[str]:
```
mitchmcdee: I think make this a util function (that goes in the util folder) that accepts a single message and returns the channel substituted message. You can just split on whitespace, convert `#something` where necessary and then rejoin

uqcsbot/scripts/jobs.py
```diff
@@ -21,68 +23,86 @@
...
     jobs_bulletin = chan
     jobs_discussion = bot.channels.get("jobs-discussion")
-    user = bot.users.get(evt.get("user"))
+    user = bot.users.get(event.get("user"))
 
     if user is None or user.is_bot:
         return
 
-    bot.post_message(jobs_bulletin, f"{user.display_name} has posted a new job in <#{jobs_bulletin.id}>! "
+    bot.post_message(jobs_bulletin, f"{user.display_name} has posted a new job! "
```
mitchmcdee: small request, could you fix `display_name` so that it defaults to the user's `real_name`? you'll have to do this in `api.py`. lmk if you need help with that

mitchmcdee: oh yeah, reason is that `display_name` is optional for older accounts (such as mine I believe?)

mitchmcdee: alternatively, you could set `name = display_name or real_name` in the user class of api.py which would make things easier, and then change all usages of it throughout the codebase (where applicable)

mitchmcdee: up to you

mitchmcdee: also, did we want to do a double message here? i.e. send to jobs-discussion too?

nicklambourne:
Have addressed these issues in #442.

PR #380, opened 2018-09-18
Fixed !whatsdue for semester 2 (fixes #369)

### LINKED ISSUE #369: Fix !whatsdue for semester 2
When using !whatsdue, if a course was available in semester 1, it will display semester 1 assessment info rather than semester 2 info

KentonLam:
Computes the hex offering string to use in get_course_profile_url so the current semester is highlighted and parsed.

In June and earlier, gets semester 1, otherwise gets semester 2. This is the same as get_current_exam_period so it should be fine..? I couldn't really test this because #369 only occurred when both semesters 1 and 2 were listed under "Current course offerings" during the first couple weeks of sem 2.

Also, it would be nice to have an argument in !whatsdue for which semester and campus to query.

### DIFF
--- a/uqcsbot/utils/uq_course_utils.py
+++ b/uqcsbot/utils/uq_course_utils.py
@@ -9,7 +9,7 @@ from binascii import hexlify
 BASE_COURSE_URL = 'https://my.uq.edu.au/programs-courses/course.html?course_code='
 BASE_ASSESSMENT_URL = 'https://www.courses.uq.edu.au/student_section_report.php?report=assessment&profileIds=' # noqa
 BASE_CALENDAR_URL = 'http://www.uq.edu.au/events/calendar_view.php?category_id=16&year='
-OFFERING_QUERY = '&offer='
+OFFERING_PARAMETER = 'offer'
 
 
 class DateSyntaxException(Exception):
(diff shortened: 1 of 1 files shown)


nicklambourne:
Hey @KentonLam, can you chuck up some screenshots of this working in the uqcs testing slack and I'll merge this? (apologies for the delay).

uqcsbot/utils/uq_course_utils.py
```diff
@@ -52,12 +54,27 @@ def __init__(self, url, status_code):
...
+    location = 'IN' if is_internal else 'EX'
+    return hexlify(f'{campus}{semester}{location}'.encode('utf-8')).decode('utf-8')
+
 
 def get_course_profile_url(course_name):
     '''
     Returns the URL to the latest course profile for the given course.
     '''
-    course_url = BASE_COURSE_URL + course_name
+    course_url = BASE_COURSE_URL + course_name + OFFERING_QUERY + get_offering_code()
```
mitchmcdee: Since we now have more than one param, I would suggest using requests `params` as seen here http://docs.python-requests.org/en/master/user/quickstart/#passing-parameters-in-urls :^)

mitchmcdee:
Could you also fix #381 in this PR?

KentonLam:
image
@nicklambourne Screenshot, but you can't actually reproduce #369 anymore. It only happened when sem 1 and 2 were under "Current course offerings".

@mitchmcdee Done and done 👍

PR #388, opened 2018-10-28
added xkcd

nicklambourne:
screen shot 2018-10-29 at 12 45 11 am


uqcsbot/scripts/xkcd.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def get_by_id(comic_number: int) -> str:
+    """
+    Gets an xkcd comic based on its unique ID/sequence number.
+    :param comic_number: the ID number of the xkcd comic to retrieve.
+    :return: a response containing either a comic URL or an error message.
+    """
+    if comic_number < 0:
+        return "Invalid xkcd ID, it must be a positive integer."
```
mitchmcdee: `< 0` implies non-negative, should it be `<=`?

uqcsbot/scripts/xkcd.py
```diff
@@ -0,0 +1,75 @@
...
+    url = f"{XKCD_BASE_URL}{str(comic_number)}"
+    response = requests.get(url)
+    if response.status_code == 200:
```
mitchmcdee: would prefer if this was flipped to:

```
if response.status_code != 200:
    return "Could not retrieve an xkcd with that ID (are there even that many?)"
return url
```

mitchmcdee: but that's just personal style preference, up to you

uqcsbot/scripts/xkcd.py
```diff
@@ -0,0 +1,75 @@
...
+def get_by_id(comic_number: int) -> str:
+    """
+    Gets an xkcd comic based on its unique ID/sequence number.
+    :param comic_number: the ID number of the xkcd comic to retrieve.
+    :return: a response containing either a comic URL or an error message.
+    """
+    if comic_number < 0:
+        return "Invalid xkcd ID, it must be a positive integer."
+    url = f"{XKCD_BASE_URL}{str(comic_number)}"
+    response = requests.get(url)
```
mitchmcdee: might be more appropriate to use request params here, but up to you

uqcsbot/scripts/xkcd.py
```diff
@@ -0,0 +1,75 @@
...
+
+def get_by_search_phrase(phrase: str) -> str:
+    """
+    Uses the site relevantxkcd.appspot.com to identify the most appropriate xkcd comic
+    based on the phrase provided.
+    :param phrase: the phrase to find an xkcd comic related to.
+    :return: the URL of the most relevant comic for that search phrase.
+    """
+    url = f"{RELEVANT_XKCD_URL}{quote(phrase)}"
+    response = requests.get(url)
```
mitchmcdee: might be more appropriate to use request params here, but up to you

uqcsbot/scripts/xkcd.py
```diff
@@ -0,0 +1,75 @@
...
+    relevant_comics = response.content.decode().split("\n")[2:]
```
mitchmcdee: nit: maybe provide a comment on what's going on over the next few lines-- maybe a snippet example?

uqcsbot/scripts/xkcd.py
```diff
@@ -0,0 +1,75 @@
...
+@loading_status
+def handle_xkcd(command: Command) -> None:
+    """
+    `!xkcd [COMIC_ID|SEARCH_PHRASE]` - Returns the xkcd comic associated
+    with the given COMIC_ID (an integer) or matching the SEARCH_PHRASE.
+    Providing no arguments will return the most recent comic.
+    """
+    if command.has_arg():
+        argument = command.arg
+        try:
```
mitchmcdee: would prefer if this was abstracted into an "is_id" function or something

i.e.
```
def is_id(argument):
    try:
        int(argument)
    except ValueError:
        return False
    else:
        return True
```

uqcsbot/scripts/xkcd.py
mitchmcdee: so that sounds like the fail case should be `<= 0` then? :P my comment was more-so to do with the statement "it must be a positive integer", of which 0 is not ofc. Unless i'm missing something I think this should be `<= 0`

nicklambourne: Right, my bad. I completely misunderstood what you were getting at. Fixed.

uqcsbot/scripts/xkcd.py
nicklambourne: I think I've addressed this, let me know if what I added is all g.

uqcsbot/scripts/xkcd.py
nicklambourne: Is there a way do this when it's not a query? Final url looks like https://xkcd.com/301, I couldn't see a way to do that in requests with params.

mitchmcdee: ah fair, ignore my comment

mitchmcdee:
lgtm, merge whenever you're ready

PR #397, opened 2018-12-01
[WIP] Add '!advent' command

ldgrp:
A session id is required to access the Advent of Code private leaderboard. I think uqcsbot should have its own account just so it can have access to the API.

Any ideas?

### DIFF
--- a/test/test_advent.py
+++ b/test/test_advent.py
@@ -1,10 +1,9 @@
+import os
 import requests
 
-# TODO: Move the LEADERBOARD_CODE and SESSION_ID someplace else to miniize duplication
-LEADERBOARD_URL = 'https://adventofcode.com/2018/leaderboard/private/view/'
-LEADERBOARD_CODE = '246889'
-# Session ID goes here
-SESSION_ID = ''
+# TODO: Move the LEADERBOARD_URL and SESSION_ID someplace else to miniize duplication
+LEADERBOARD_URL = 'https://adventofcode.com/2018/leaderboard/private/view/246889.json'
+SESSION_ID = os.environ['AOC_SESSION_ID']
 
 
 def test_advent_normal():
@@ -12,8 +11,7 @@ def test_advent_normal():
     Tests that the private leaderboard can be accessed and the
     JSON structure can be parsed
     """
-    url = "{}{}.json".format(LEADERBOARD_URL, LEADERBOARD_CODE)
-    response = requests.get(url, cookies={"session": SESSION_ID})
+    response = requests.get(LEADERBOARD_URL, cookies={"session": SESSION_ID})
 
     assert response.status_code == 200
     assert response.json()
--- a/test/test_advent.py
+++ b/test/test_advent.py
@@ -1,9 +1,5 @@
-import os
 import requests
-
-# TODO: Move the LEADERBOARD_URL and SESSION_ID someplace else to miniize duplication
-LEADERBOARD_URL = 'https://adventofcode.com/2018/leaderboard/private/view/246889.json'
-SESSION_ID = os.environ['AOC_SESSION_ID']
+from uqcsbot.scripts import advent
 
 
 def test_advent_normal():
@@ -11,7 +7,7 @@ def test_advent_normal():
     Tests that the private leaderboard can be accessed and the
     JSON structure can be parsed
     """
-    response = requests.get(LEADERBOARD_URL, cookies={"session": SESSION_ID})
+    response = requests.get(advent.LEADERBOARD_URL, cookies={"session": advent.SESSION_ID})
 
     assert response.status_code == 200
     assert response.json()
--- a/test/test_advent.py
+++ b/test/test_advent.py
@@ -1,0 +1,20 @@
+import requests
+
+# TODO: Move the LEADERBOARD_CODE and SESSION_ID someplace else to miniize duplication
+LEADERBOARD_URL = 'https://adventofcode.com/2018/leaderboard/private/view/'
+LEADERBOARD_CODE = '246889'
+# Session ID goes here
+SESSION_ID = ''
+
+
+def test_advent_normal():
+    """
+    Tests that the private leaderboard can be accessed and the
+    JSON structure can be parsed
+    """
+    url = "{}{}.json".format(LEADERBOARD_URL, LEADERBOARD_CODE)
+    response = requests.get(url, cookies={"session": SESSION_ID})
+
+    assert response.status_code == 200
+    assert response.json()
+    assert response.json()['members']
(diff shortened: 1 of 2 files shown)


ldgrp:
advent output
image

test/test_advent.py
```diff
@@ -0,0 +1,20 @@
...
+# Session ID goes here
+SESSION_ID = ''
+
+
+def test_advent_normal():
+    """
+    Tests that the private leaderboard can be accessed and the
+    JSON structure can be parsed
+    """
+    url = "{}{}.json".format(LEADERBOARD_URL, LEADERBOARD_CODE)
```
nicklambourne: This can be replaced with an f-string e.g.:
```python
url=f"{LEADERBOARD_URL}{LEADERBOARD_CODE}.json"
```

test/test_advent.py
```diff
@@ -0,0 +1,20 @@
+import requests
+
+# TODO: Move the LEADERBOARD_CODE and SESSION_ID someplace else to miniize duplication
```
nicklambourne: You should, hypothetically be able to import LEADERBOARD_CODE from the actual script.

uqcsbot/scripts/advent.py
```diff
@@ -0,0 +1,61 @@
...
+    Post the Advent of Code Leaderboard on #contests
+    '''
+    channel = bot.channels.get("contests")
+    leaderboard = get_leaderboard()
+    members = get_members(leaderboard['members'])
+
+    message = "```\n"
+    message += "Score Stars Name\n"
+    for member in members:
+        message += "{:5} {:5} {}\n".format(member.score, member.stars, member.name)
```
nicklambourne: f-string

uqcsbot/scripts/advent.py
```diff
@@ -0,0 +1,61 @@
...
+        self.stars = stars
+
+    def __lt__(self, other):
+        return self.score > other.score or self.stars > other.stars
+
+#@bot.on_schedule('cron', hour=15, timezone='Australia/Brisbane')
+@bot.on_command("advent")
+def advent(command: Command) -> None:
+    '''
+    Post the Advent of Code Leaderboard on #contests
```
nicklambourne: Need proper usage docstring to be picked up by !help e.g.:
```
"""
`!xkcd [COMIC_ID|SEARCH_PHRASE]` - Returns the xkcd comic associated with the given COMIC_ID (an integer) or matching the SEARCH_PHRASE. Providing no arguments will return the most recent comic.
"""
```

uqcsbot/scripts/advent.py
```diff
@@ -0,0 +1,61 @@
...
+
+    members.sort()
+    return members
+
+def get_leaderboard() -> Dict:
+    '''
+    Returns a json dump of the leaderboard
+    '''
+    try:
+        url = "{}{}.json".format(LEADERBOARD_URL, LEADERBOARD_CODE)
```
nicklambourne: f-string

uqcsbot/scripts/advent.py
```diff
@@ -0,0 +1,61 @@
+from uqcsbot import bot, Command
+from requests.exceptions import RequestException
+from typing import Dict, List
+import requests
+
+LEADERBOARD_URL = 'https://adventofcode.com/2018/leaderboard/private/view/'
+LEADERBOARD_CODE = '246889'
+# Session ID goes here
+SESSION_ID = ''
```
nicklambourne: This can be stored in an environment variable, e.g.:
os.environ['AOC_SESSION_ID']
I can modify the supervisord config to store a session ID, though I expect they will expire at some point (though maybe not until AoC finishes)

test/test_advent.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    Tests that the private leaderboard can be accessed and the
+    JSON structure can be parsed
+    """
+    url = "{}{}.json".format(LEADERBOARD_URL, LEADERBOARD_CODE)
+    response = requests.get(url, cookies={"session": SESSION_ID})
+
+    assert response.status_code == 200
+    assert response.json()
+    assert response.json()['members']
```
nicklambourne: Might be worth testing your sort given that you're overriding default behaviour. Up to you.

uqcsbot/scripts/advent.py
```diff
@@ -0,0 +1,61 @@
+from uqcsbot import bot, Command
+from requests.exceptions import RequestException
+from typing import Dict, List
+import requests
+
+LEADERBOARD_URL = 'https://adventofcode.com/2018/leaderboard/private/view/'
+LEADERBOARD_CODE = '246889'
```
nicklambourne: Given the fact that the URL will have to be updated manually every year anyway (and you never use it on its own), is there any point in having these two as separate variables?

uqcsbot/scripts/advent.py
ldgrp: oops. didn't realize you can do all the str.format() stuff with f-strings :laughing:

ldgrp:
^ oops :open_mouth:

ldgrp:
Thanks for the feedback @nicklambourne. I _think_ i've incorporated most of the changes requested except for having the Leaderboard URL in a config file and maybe a better usage docstring.

nicklambourne:
@ldgrp for some reason the environment variables aren't getting picked up in the tests. Will have to look into the cause. Any idea @TRManderson ?

TRManderson:
The issue is likely the quotation marks in the env var string - I have removed these. Also, I think having the env var set for tests is a bad idea - we should instead mock the request.

Right now the session ID is public on Jenkins because it was inserted into the test run.

uqcsbot/scripts/advent.py
```diff
@@ -0,0 +1,61 @@
...
+        message += "{:5} {:5} {}\n".format(member.score, member.stars, member.name)
+    message += "```"
+
+    bot.post_message(command.channel_id, message)
+
+def get_members(members_json: Dict) -> List[Member]:
+    '''
+    Returns a sorted list of Members in the leaderboard
+    '''
+    members = []
```
MaxwellBo: `members = sorted(Member(m['name'], m['local_score'], m['stars']) for m in members_json.values())`

nicklambourne:
@ldgrp I agree with @TRManderson, we can just provide mock input for the test and avoid the underlying issue. Sorry this has taken so long to address, did you want to amend the test? Might also be worth amending the cron to only run until the 25th December 😂.

nicklambourne:
Interest in this seems to have flagged, closing for now.

kentonlam:
I'll restart work on this. Hopefully it gets merged soon before it becomes irrelevant for another year.

kentonlam:
Closing again in favour of #558. Many thanks for your work.

Issue #394, opened 2018-11-09
context for xkcd post


Guthers:
The xkcd post just sort of appears in general every now and then. It would be nice for uqcsbot to accompany it with some context such as
>"xkcd has just released a new comic"

potentially even as an embedded link in the text.

Think it might be a nice easy one at some point

PR #399, opened 2018-12-29
#394 context for XKCD comic

Paul-Haley:
As requested in issue #394, some context to the latest XKCD comics being posted. 

First PR **ever**, apologies if I have broken protocol


jack775544:
Code looks fine to me, only thing is that there should be a fallback case if the day does not happen to be on one of those 3. Even though they should always been on those days becasue of the cron job it is better to be safe then sorry.

PR #405, opened 2019-02-19
Add hint to download Slack app

MaxwellBo:
This will increase engagement, as most people, after joining, never log back on, mainly because they used the web client. 

I'm open to better copy, suggest better versions and those with the most likes, I'll amend the commit.


jdcaperon:
lgtm :+1:

PR #413, opened 2019-03-15
Adds a !crates command

mubiquity:
Adds a command !crates which does a whole bunch of stuff related to crates.io.

To summarise, it can search crates by exact name or it can return a list of crates narrowed down by a search term, category and author. It can also return a list of all category names and give more in depth detail about a category. Finally, it can return publicly available information about a user. Examples of all of these scenarios below.

### DIFF
--- a/uqcsbot/scripts/crates.py
+++ b/uqcsbot/scripts/crates.py
@@ -549,7 +549,10 @@ def handle_categories_route(channel: Channel, args: CategorySearch):
 
 
 def get_user(channel: Channel, username: str) -> Optional[UserResult]:
-    """Gets a UserResult by querying the api for the given username. None on error."""
+    """
+    Gets a UserResult by querying the crates.io api for the given username.
+    None on error.
+    """
     url = f'{BASE_URL}/users/{username}'
     response = requests.get(url)
 
--- a/uqcsbot/scripts/crates.py
+++ b/uqcsbot/scripts/crates.py
@@ -1,5 +1,6 @@
 import argparse
 import json
+from enum import Enum
 from typing import NamedTuple, Union, Optional, List, Dict, Tuple
 
 import requests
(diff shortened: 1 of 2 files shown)


uqcsbot/scripts/crates.py
```diff
@@ -0,0 +1,561 @@
...
+
+    # Get the category names
+    categories = [cat.get('name') if 'name' in cat else cat.get('id', '') for cat in raw_categories]
+
+    return categories, total
+
+
+def display_all_categories(channel: Channel, args: CategorySearch):
+    """Displays just the names of all the categories in one big list"""
+    cats, tot = get_category_page(channel, args.sort, 1)
```
nicklambourne: Please avoid using shorthand for variables where possible, it reduces clarity and we'd prefer comprehensibility over concision.

nicklambourne: Please avoid using shorthand for variables where possible, it reduces clarity and we'd prefer comprehensibility over concision. Common shorthand like 'args' for arguments is okay (given its recurrent use), but 'tot' and 'cats' is confusing.

uqcsbot/scripts/crates.py
```diff
@@ -0,0 +1,561 @@
...
+                'text': f'```{category_string}```'
+            }
+        },
+    ]
+
+    bot.post_message(channel, '', blocks=blocks)
+
+
+def display_specific_category(channel: Channel, args: CategorySearch):
+    """Displays a single category in more detail"""
```
nicklambourne: Category of what?

uqcsbot/scripts/crates.py
```diff
@@ -0,0 +1,561 @@
...
+
+    name = raw_category.get('name')
+    name = raw_category.get('id') if name is None else name
+    desc = raw_category.get('description', 'No description provided')
+
+    category = CategoryResult(name, desc, raw_category['crates_cnt'])
+
+    # Format the message
+    blocks = [
+        {
```
nicklambourne: These dictionaries look like the kind of thing that could be produced algorithmically. Just a thought...

uqcsbot/scripts/crates.py
```diff
@@ -0,0 +1,561 @@
...
+
+def handle_categories_route(channel: Channel, args: CategorySearch):
+    """Handles the categories sub-command by determining whether or not to display all categories or just one"""
+    if args.name:
+        display_specific_category(channel, args)
+    else:
+        display_all_categories(channel, args)
+
+
+def get_user(channel: Channel, username: str) -> Optional[UserResult]:
```
nicklambourne: I don't know if it's just me, but it took a while to work out that this was referring to crates.io users, not slack ones. Could it be made more clear?

uqcsbot/scripts/crates.py
mubiquity: I have added some classes and functions to do this in a rather simplistic way however I feel as though it doesn't really belong in a script file.
I think the bot might benefit from having something dedicated to constructing responses using the new blocks layout as it does seem to fit well with modular message construction.

nicklambourne: You're right, we should push this upwards. Want to make an issue and link back to this?

PR #420, opened 2019-04-28
Create dice.py

bradleysigma:


uqcsbot/scripts/dice.py
```diff
@@ -0,0 +1,31 @@
+from random import choice
+from uqcsbot import bot, Command
+
+def is_number(message):
```
nicklambourne: Python strings have a builtin str.isnumeric() method that can replace this whole method.

uqcsbot/scripts/dice.py
```diff
@@ -0,0 +1,31 @@
...
+    '''
+    `!dice [number]` - Rolls 1 or more six sided dice (d6).
+    '''
+    if command.has_arg() and is_number(command.arg):
+        num = max(int(command.arg), 1)
+    else:
+        num = 1
+
+    resp = []
+    emoj = (':dice-one:', ':dice-two:', ':dice-three:', ':dice-four:', ':dice-five:', ':dice-six:')
```
nicklambourne: No shorthand please (e.g. emoj instead of emoji, resp instead of response), makes the code less readable.

Could num be called rolls, or something else more descriptive.

uqcsbot/scripts/dice.py
```diff
@@ -0,0 +1,31 @@
...
+        return False
+    return True
+
+
+@bot.on_command("dice")
+def handle_dice(command: Command):
+    '''
+    `!dice [number]` - Rolls 1 or more six sided dice (d6).
+    '''
+    if command.has_arg() and is_number(command.arg):
```
nicklambourne: command.arg.isnumeric()

uqcsbot/scripts/dice.py
```diff
@@ -0,0 +1,31 @@
+from random import choice
+from uqcsbot import bot, Command
+
+def is_number(message):
+    '''
```
nicklambourne: PEP8 suggests triple double quotes for docstrings.

uqcsbot/scripts/dice.py
bradleysigma: This was copied from `mock.py`. (`isnumeric()` returns `False` if string contains a `.`.)  
Removed.

uqcsbot/scripts/dice.py
bradleysigma: Fixed.

uqcsbot/scripts/dice.py
bradleysigma: `command.has_arg()` needed to short circuit in the case of no arguments.  
Fixed.

uqcsbot/scripts/dice.py
bradleysigma: This was copied from `mock.py`.

bradleysigma:
image

PR #421, opened 2019-04-29
Change the interval between reactions from 1 second to 2 seconds

mubiquity:
Sorry about this. From testing it seemed like 1 second was enough (does seem to work the majority of the time but not always). 

This changes it from 1 second to 2 seconds which should always be enough (hopefully). If not, I won't increase it more as that would be a trade off for usability.

### DIFF
--- a/uqcsbot/scripts/trivia.py
+++ b/uqcsbot/scripts/trivia.py
@@ -24,7 +24,7 @@ QuestionData = NamedTuple('QuestionData',
 ReactionUsers = NamedTuple('ReactionUsers', [('name', str), ('users', Set[str])])
 
 # Customisation options
-REACT_INTERVAL = 1  # The interval between reactions being made for the possible answers (prevents order changing)
+REACT_INTERVAL = 2  # The interval between reactions being made for the possible answers (prevents order changing)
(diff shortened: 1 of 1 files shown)


TRManderson:
I give you one (1) very large yike for relying on timing to order stuff

mubiquity:
Yeah, but as far as I know there is no way around that? Slack determines reaction order by their timestamps which only go to the nearest second. If there is a better way let me know.

TRManderson:
Voteythumbs is always in the correct order: https://github.com/UQComputingSociety/uqcsbot/blob/master/uqcsbot/scripts/voteythumbs.py

mubiquity:
Yep, not gonna lie, that confused me. I looked at that when I was having this problem originally but I have no idea that works and !trivia doesn't.

TRManderson:
I suspect the scheduler is to blame

mubiquity:
It still happened with non scheduled !trivia and back when the reactions were just done in the same way as voteythumbs.

mubiquity:
Going to close this for the time being while I test for a better solution. It works reasonably well as is and isn't that big of a problem anyway.

PR #428, opened 2019-04-30
Create weather.py

bradleysigma:

PR #430, opened 2019-05-08
Create "!weather" command

bradleysigma:
See #428 
image


uqcsbot/scripts/weather.py
```diff
@@ -0,0 +1,215 @@
+from uqcsbot import bot, Command
+from urllib.request import urlopen
+import xml.etree.ElementTree as ET
+from datetime import datetime as DT
+from typing import Union, Tuple
+
+
+def get_xml(state: str) -> Union[None, ET.Element]:
```
nicklambourne: `Optional[ET.Element]`

PR #434, opened 2019-05-14
Naive fix for hoogle. Strip html and formatting

### LINKED ISSUE #433: !hoogle is broken
```
  File "/srv/uqcsbot/lib/python3.6/site-packages/uqcsbot/scripts/hoogle.py", line 56, in handle_hoogle
    results = json.loads(http_response.content).get('results', [])
AttributeError: 'list' object has no attribute 'get'
```
Seems like the response format changed

jdcaperon:
Fixes #433

For reference, the response format directly returns a list of dicts now. Example dict:
```
{"url":"https://hackage.haskell.org/package/base/docs/Prelude.html#v:const","module":{"url":"https://hackage.haskell.org/package/base/docs/Prelude.html","name":"Prelude"},"package":{"url":"https://hackage.haskell.org/package/base","name":"base"},"item":"<span class=name><s0>const</s0></span> :: a -&gt; b -&gt; a","type":"","docs":"<tt>const x</tt> is a unary function which evaluates to <tt>x</tt> for\nall inputs.\n\n<pre>\n&gt;&gt;&gt; const 42 \"hello\"\n42\n</pre>\n\n<pre>\n&gt;&gt;&gt; map (const 42) [0..3]\n[42,42,42,42]\n</pre>\n"}
```

The responses now also contain HTML markup. It's trivial to strip from the type sig but less so from the docs segment without the formatting for the docs section being a little awkward.

image

image

### DIFF
--- a/uqcsbot/scripts/hoogle.py
+++ b/uqcsbot/scripts/hoogle.py
@@ -1,4 +1,5 @@
 from uqcsbot import bot, Command
+from uqcsbot.utils.command_utils import loading_status
 import requests
 import json
 import html
@@ -22,6 +23,7 @@ def pretty_hoogle_result(result: dict, is_verbose: bool) -> str:
 
 
 @bot.on_command("hoogle")
+@loading_status
 def handle_hoogle(command: Command):
     '''
     `!hoogle [-v] [--verbose] <TYPE_SIGNATURE>` - Queries the Hoogle Haskell API search engine,
--- a/uqcsbot/scripts/hoogle.py
+++ b/uqcsbot/scripts/hoogle.py
@@ -2,7 +2,7 @@ from uqcsbot import bot, Command
 import requests
 import json
 import html
-
+import re
 
 def get_endpoint(type_sig: str) -> str:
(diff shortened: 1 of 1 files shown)


uqcsbot/scripts/hoogle.py
```diff
@@ -11,9 +12,9 @@ def get_endpoint(type_sig: str) -> str:
 
 
 def pretty_hoogle_result(result: dict, is_verbose: bool) -> str:
-    url = result['location']
-    type_sig = result['self']
-    docs = result['docs']
+    url = result['url']
+    type_sig = re.sub('<[^<]+?>', '', result['item'])
+    docs = re.sub('<[^<]+?>', '', result['docs']).replace('\n',' ').replace('&gt;&gt;&gt;','\n>')
```
nicklambourne: Is this supposed to have a > in the last replace?
`'\n>'`

jdcaperon: Yeah, this allows the code to be still "quoted" in the slack message, alternatively, it can be removed so that the code isn't quoted.

PR #438, opened 2019-05-25
Include ITEE Seminar information in !events command.

patrickmeiring:
As discussed in #uqcs-meta. This change adds seminars run by the School of ITEE to the !events command.
Seminar information is sourced by scraping the https://www.itee.uq.edu.au/seminar-list page on the school website.

Comments:
- Because information about the speaker is not on the main event listings page, I am having to visit the linked event details pages. As the number of page visits required scales based on the number of events, this may have adverse impacts on performance. On my system the command doesn't take more than a second or two, but I suggest we monitor this and if it becomes a problem, we investigate alternative solutions. These could include any of the following:
  1. Pre-emptively loading and caching speaker information, 
  2. Dropping the speaker name from the event, 
  3. Turning the script into a 'push' system, rather than a 'pull' system, where event information is automatically pushed into the #events channel on a periodic basis/when it changes.
- In addition to adding tests for my changes, I have taken the liberty of adding some tests for the other features of the !events command
- As this is the first time I'm contributing to something written in python, you might want to review my code carefully... I did not know python before today.


uqcsbot/utils/itee_seminar_utils.py
```diff
@@ -0,0 +1,133 @@
...
+    code.
+    """
+    def __init__(self, url: str, status_code: int):
+        self.message = f'Received status code {status_code} from \'{url}\'.'
+        self.url = url
+        self.status_code = status_code
+        super().__init__(self.message, self.url, self.status_code)
+
+
+def get_seminars():
```
nicklambourne: Compound return types e.g. List[string] can be achieved by importing List etc. from the standard library module `typing`.

uqcsbot/utils/itee_seminar_utils.py
```diff
@@ -0,0 +1,133 @@
...
+    Returns the content of the page summarising upcoming seminars.
+    :return: The HTML of the page containing upcoming seminar information.
+    """
+    http_response = requests.get(ITEE_SEMINAR_LIST_URL)
+    if http_response.status_code != requests.codes.ok:
+        raise HttpException(ITEE_SEMINAR_LIST_URL, http_response.status_code)
+    return http_response.content
+
+
+def get_seminar_summary(seminar_row):
```
nicklambourne: Return type annotation. `from typing import Tuple`

I won't add comments for the other issues of this type here.

uqcsbot/utils/itee_seminar_utils.py
```diff
@@ -0,0 +1,133 @@
...
+    Returns the content of the given seminar details page. Stubbed in unit tests.
+    :return: The HTML of the page containing seminar details.
+    """
+    http_response = requests.get(seminar_url)
+    if http_response.status_code != requests.codes.ok:
+        raise HttpException(seminar_url, http_response.status_code)
+    return http_response.content
+
+
+def parse_seminar_date(date_string: str, url: str):
```
nicklambourne: `from typing import Optional`

`-> Optional[datetime]`

test/test_events.py
```diff
@@ -0,0 +1,229 @@
...
+NO_EVENTS_MESSAGE = "_There don't appear to be any events in the next *2* weeks_\r\n" \
+    "For a full list of events, visit: https://uqcs.org.au/calendar.html " \
+    "and https://www.itee.uq.edu.au/seminar-list"
+
+
+def mocked_html_summary_get_typical():
+    """
+    Returns locally stored HTML that represents a typical seminar listing.
+    """
+    f = open("test/ITEE_Upcoming_Seminars.html", "r")
```
nicklambourne: Nitpicking (probably too much 2002 marking), but maybe a variable name like html_data, or just return the chained function calls. Not super worried about whether this/others like this are actioned.

patrickmeiring:
Thanks all for the feedback. @TRManderson @nicklambourne 

I basically haven't written any python before, so it's good to learn how I can add those type annotations I was missing. Also thanks for the pointer on File I/O -- I will have a look.

I will address these and come back with an updated PR in a few days when I have time.

nicklambourne:
```
with open('filename.html') as html_data:
    html_data.do_thing()
```

patrickmeiring:
@TRManderson @nicklambourne Code changes made. Please review the changes I've made and merge if you are happy.

PR #435, opened 2019-05-17
Create !crisis, !emergency and !mentalhealth

d-lord:
Both do the same thing - provide a list of crisis resources for quick reference.

Happy to add any other resources that may be of use :)

PR #441, opened 2019-05-26
Create !crisis, !mentalhealth and !emergency commands

d-lord:
Re-merge of #435 after Nick did some stuff with the tests. Content is identical.

All are aliases for the same thing - provide a list of crisis resources for
quick reference.


d-lord:
Thanks for the reviews everyone, and thanks @nicklambourne for fixing the tox problem! 😄

PR #446, opened 2019-06-08
Emojify

bradleysigma:
Long messages are now no longer split in the middle of an emoji. 
(Also, `:office_excel:` is now an `x`, not an `e`.)


bradleysigma:
> Can you add a total length cap at something reasonable as well?

Is 3991 characters reasonable? If so, done.

nicklambourne:
> > Can you add a total length cap at something reasonable as well?
> 
> Is 3991 characters reasonable? If so, done.

Um, no. 300-500 absolute tops.

bradleysigma:
Now limits to one message (4000 characters) from UQCSbot, with ellipsis. Also preserves linebreaks.

PR #476, opened 2019-08-16
fix for doubling of xkcd announcements

nicklambourne:
Uses cron more effectively to avoid duplicate posts of the same xkcd

### DIFF
--- a/uqcsbot/scripts/xkcd.py
+++ b/uqcsbot/scripts/xkcd.py
@@ -1,6 +1,6 @@
-import datetime
 import requests
-import feedparser
+from feedparser import FeedParserDict, parse
+from datetime import datetime, timedelta
 from urllib.parse import quote
 from uqcsbot import bot, Command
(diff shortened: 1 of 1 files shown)


bradleysigma:
If I understand this correctly, this now check at five minutes past every hour (on Mon, Wed Fri) if a comic has been released in the past five minutes, and if so, posts it. Is this correct?

uqcsbot/scripts/xkcd.py
```diff
@@ -89,7 +89,7 @@ def handle_xkcd(command: Command) -> None:
     bot.post_message(command.channel_id, response, unfurl_links=True, unfurl_media=True)
 
 
-@bot.on_schedule('cron', hour=14, minute=1, day_of_week='mon,wed,fri',
+@bot.on_schedule('cron', minute=5, day_of_week='mon,wed,fri',
```
bradleysigma: This will run the function whenever the minutes is 5 i.e. five past every hour. Probably not what you want.

PR #477, opened 2019-08-18
Create `!scry`

bradleysigma:
Uses the Scryfall api to get an image of the requested _Magic: the Gathering_ card.
image


bradleysigma:
Inspiration
image

PR #483, opened 2019-10-01
Adding daily "on this day" using channel pins

bradleysigma:
image
This will, once a day, find a pin that was posted on the same date in a previous year*, and repost it in the same channel.


kumar-ish:
lgtm

PR #492, opened 2019-10-18
Add source filter to `!events`

bradleysigma:
image


kumar-ish:
Lgtm. We may want to make the default only UQCS events since that's what 95% of the people use it for, and the seminars crowd it up quite a bit, but I'll do that myself at a later date (along with some other minor changes).

PR #499, opened 2019-11-14
Add background to latex images

jsutton101:
Changed background of latex images from transparent to white so that they can be viewed on dark mode


kumar-ish:
Lgtm.

PR #503, opened 2020-01-02
Slackclient v2 full update

TRManderson:
This updates our slack client and testing code to the latest version. I've also taken the liberty of prepping uqcsbot for the "two different token" aka "granular scopes" update that destroyed it last time.

Right now we're having to use an old-style token spun up specifically but if they ever reenable bot users we'll probably have to do something different there. If they never do, we should rewrite to use the events API.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -2,7 +2,6 @@
 Configuration for Pytest
 """
 
-from unittest.mock import MagicMock
 from itertools import islice
 from functools import partial
 from collections import defaultdict
@@ -11,7 +10,7 @@ from typing import Optional
 import pytest
 from slack import WebClient
 import uqcsbot as uqcsbot_module
-from uqcsbot.api import APIWrapper, APIMethodProxy
+from uqcsbot.api import APIWrapper
 from uqcsbot.base import UQCSBot, Command
 from copy import deepcopy
 
@@ -158,7 +157,8 @@ class MockUQCSBot(UQCSBot):
             return channel.get(channel_type, False)
 
         if channel_type == 'all':
-            filter_function = lambda x: True
+            def filter_function(*args):
+                return True
             channel_type = 'channels'
         elif channel_type == 'channels':
             filter_function = partial(is_channel_type, channel_type='is_public')
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -9,9 +9,9 @@ from collections import defaultdict
 import time
 from typing import Optional
 import pytest
-from slackclient import SlackClient
+from slack import WebClient
 import uqcsbot as uqcsbot_module
-from uqcsbot.api import APIWrapper
+from uqcsbot.api import APIWrapper, APIMethodProxy
 from uqcsbot.base import UQCSBot, Command
 from copy import deepcopy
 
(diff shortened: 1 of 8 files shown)


TRManderson:
Pls wait for tests and at least one review

test/conftest.py
```diff
@@ -41,6 +41,10 @@
                      'is_private': True, 'is_user_deleted': False,
                      'user': TEST_USER_ID}
 }
+for item in ['is_im', 'is_public', 'is_private', 'is_group']:
+    for chan in TEST_CHANNELS.values():
```
BraeWebb: chan as a variable name 🤢 kids and their abbreviations these days

TRManderson:
At this point I no longer have any idea whether or not this will work

TRManderson:
yeet to prod successful

uqcsbot/api.py
```diff
@@ -56,16 +63,27 @@ def __call__(self, **kwargs) -> dict:
 
         Attempts to retry the API call if rate-limited.
         """
-        fn = partial(self._client.api_call, self._method, **kwargs)
+        def get_fn(call_type):
+            client = getattr(self, f'_{call_type}_client')
```
BlueDragon23: Will this always be exactly user client or bot client? I'm not a python main, but this seems more magic than it needs to be

uqcsbot/api.py
```diff
@@ -56,16 +63,27 @@ def __call__(self, **kwargs) -> dict:
...
+            except slack.errors.SlackApiError as e:
+                result = e.response
             if not result['ok'] and result['error'] == 'ratelimited':
                 retry_after_secs = int(result['headers']['Retry-After'])
                 LOGGER.info(f'Rate limited, retrying in {retry_after_secs} seconds')
                 time.sleep(retry_after_secs)
                 retry_count += 1
+            elif not result['ok'] and result['error'] == 'not_allowed_token_type':
+                call_type = {'bot': 'user', 'user': 'bot'}[call_type]
+                retry_count += 1
```
BlueDragon23: So this could switch back and forth between bot and user accounts 5 times? It won't break anything, but it doesn't seem ideal. Can we fail out of this if we've already tried both accounts?

uqcsbot/api.py
```diff
@@ -56,16 +63,27 @@ def __call__(self, **kwargs) -> dict:
...
+        def get_fn(call_type):
+            client = getattr(self, f'_{call_type}_client')
+            method = getattr(client, self._method.replace('.', '_'))
+            return partial(method, **kwargs)
         retry_count = 0
+        call_type = _CLIENT_METHOD_REGISTRY.get(self._method, 'bot')
         while retry_count < 5:
-            result = fn()
+            try:
+                result = get_fn(call_type)()
```
BlueDragon23: So we cleverly retrieve a partially applied function so it can immediately be called each time. Can we just call the method instead of partially applying it in `get_fn`?

uqcsbot/api.py
TRManderson: Yeah this was just laziness lel

uqcsbot/api.py
TRManderson: Good catch

test/conftest.py
```diff
@@ -142,19 +156,23 @@ def is_channel_type(channel, channel_type):
             '''
             return channel.get(channel_type, False)
 
-        if channel_type == 'channels':
+        if channel_type == 'all':
+            def filter_function(*args):
+                return True
+            channel_type = 'channels'
```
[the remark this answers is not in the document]
TRManderson: Due to the way the channels are generated, there's definitely a better approach here, thanks

Issue #493, opened 2019-11-06
Commands in threads causes the bot to respond in the channel proper


bradleysigma:
Ideally, the response would match the command e.g. comments posted only in the thread would have a response only in the thread, and comments posted to both the thread and channel would have a response in both the thread and channel.

`yelling.py` has the above implemented, but ideally there would be some method to automate it for any command.

PR #505, opened 2020-01-03
UQCSBot: Ressurection

TRManderson:
This is a cleaned up version of my PR #503, which also includes some extra stuff like a partial fix for #493 (to be implemented separately per script). There's a lot going on soz. This is needed for medium term UQCSBot stability.


TRManderson:
I've verified dev mode still works without further coordination effort RE user tokens, so that's nice

PR #511, opened 2020-02-24
Updated Calendar Format to Distinguish UQCS from ITEE

nicklambourne:
Screen Shot 2020-02-24 at 9 33 34 pm

Will need to work out later if we want to check message formatting, and if so how to capture attachment output text.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,8 @@ install_requires = [
     'google-api-python-client',
     'pytest-runner',
     'unidecode',
-    'feedparser'
+    'feedparser',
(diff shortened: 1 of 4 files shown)


nicklambourne:
> LGTM, is it possible to test the event blocks output as they are sent to the channel?

I imagine so. Will have a look this evening.

nicklambourne:
Going to merge this, will evaluate testing content later.

jdcaperon:
sgtm, was more of a question than a requirement, this one is pretty easy to notice breaking in production

PR #513, opened 2020-02-28
Creates `exclamation.py`, which discourages multiple exclamation marks

bradleysigma:
image


nicklambourne:
Too niche. Could be considered offensive.

TRManderson:
Yeah that comment defs offensive

bradleysigma:
FWIW, it's a Terry Pratchett (from _~Faust~ Eric_).

Issue #411, opened 2019-03-02
Better Holidays


nicklambourne:
Add more relevant holidays to !holiday: 

- https://www.engadget.com/2016/01/01/the-complete-guide-to-geek-holidays/

These would need to be translated into a better format, but happy to stick a csv on the server or put a new table in the postgres DB to facilitate this.

I'd suggest using the geek holidays and reverting to the others if there are none on that day.

LauchieHarvey:
I scraped a website to compile a .csv. This is my first code contribution so bear with me :). Where in the directory structure should the .csv go? @nicklambourne

LauchieHarvey:
#535

nicklambourne:
Please make a new directory to put it in at `uqcsbot/uqcsbot/static`.

PR #535, opened 2020-05-24
Added extra holidays

LauchieHarvey:
This is my first pull request so bear with me :) I am not sure what branch to merge it to so I just put master :/ This contains a new `.csv` file which has geeky holidays in it. I also updated the holiday script to read from the `.csv` so that the holidays will actually show. 
I didn't know what directory was best to put the `.csv` in so I just have a constant variable in the `holiday.py` file referencing the path of the `.csv`. For now it is in the scripts directory but that can easily be changed. I tried to test it and I think from what I could tell it passed all of the tests, but I couldn't figure out how to actually get the 'holiday' message to display.


nicklambourne:
Okay, a few things:

1. Having work on the master branch of your fork of the repo is fine.
2. As I mentioned in #411, you can create a new directory at `uqcsbot/uqcsbot/static` to put the CSV in.
3. Make sure the reference path you use is relative to the uqcsbot root directory.
4. You may have to change the cron settings for the job (in the decorator) to get it to run while you're looking at it... Don't forget to test it out in our test, if you don't have access to that, DM me your email on regular Slack and I'll add you.
5. It's currently failing Jenkins checks because of error code F541: [https://jenkins.uqcs.org/job/uqcsbot-pr/555/console](https://jenkins.uqcs.org/job/uqcsbot-pr/555/console). These errors (right down the bottom) weren't caused by you, but you could fix them for us as part of the PR, or you can temporarily add --ignore=F541 to the two calls to flake8 in tox.ini and I can fix them later.

nicklambourne:
Once all of that is fixed, I'll do a proper review.

bradleysigma:
jenkins retest this please

bradleysigma:
Apparently `flake8` was updated recently, which may be the reason for the `F541` error.

LauchieHarvey:
I have made changes to the `tox.ini` (added --ignore=F541) file but there is still a failing check. I'll have to do a bit of research on the jenkins system because I've never worked with continuous integration before. @nicklambourne

kentonlam:
It appears that `--ignore` overrides the default ignore list, leading to warnings in previously-passing code. Using `--extend-ignore` instead should fix this. https://flake8.pycqa.org/en/latest/user/options.html

nicklambourne:
Looks like there's just one left:
`test/test_events.py:87:1: E302 expected 2 blank lines, found 1`
Ignore or fix, up to you.

LauchieHarvey:
After your suggestion @nicklambourne and help from @kentonlam it seems to have passed. Sorry I didn't fix the underlying issue. I need to learn more about Jenkins etc. before I have the understanding to solve it :) Thanks for your help and I'll await the PR review.

uqcsbot/scripts/holiday.py
```diff
@@ -76,6 +79,21 @@ def get_holidays_from_page(holiday_page) -> list:
     return holidays
 
 
+def get_holidays_from_csv():
```
nicklambourne, line 82: Can we get some typing on this function?

`def get_holidays_from_csv() -> List[Something]`

You may have to import complex types from the `typing` module.

LauchieHarvey: image
I'm not certain what you wanted a screenshot of. Here is the terminal output, you can see it runs successfully. The warning at the end is because I exited with a keyboard interrupt.
image
This image is from the testing slack. There are two because I ran it twice, once to check that the original website scraping holidays work and the other to test a holiday from the `geek_holidays.csv`.

LauchieHarvey: Also, see the commit below for the type hinting I added to `holiday.py`.

nicklambourne: The screenshot of it working is usually all we're after.

uqcsbot/scripts/holiday.py
```diff
@@ -58,7 +59,7 @@ def get_holiday() -> Holiday:
     return choice(holidays_today) if holidays_today else None
 
 
-def get_holidays_from_page(holiday_page) -> list:
+def get_holidays_from_page(holiday_page) -> List[Holiday]:
```
nicklambourne, line 62: Sorry this was my bad for missing it, but what exactly is `holday_page`? Will likely need a type, too.

nicklambourne:
jenkins retest this please

nicklambourne:
jenkins retest this please

PR #536, opened 2020-06-02
Emojify

bradleysigma:
Updating `!emojify` to add newly added Slack emoji.


es02:
PR pass so long as Jenkins passes :)

Issue #534, opened 2020-05-20
Don't chastise links in #yelling


mon:
Even if you put an allcaps hyperlink with an allcaps URL, slack will lowercase the domain name and the bot will yell at you about it.

d-lord:
This applies to both pasting a raw URL and using a rich text hyperlink.

PR #543, opened 2020-10-02
Updated yelling.py

paulrusthewalrus:
This change should stop the bot from chastising you just because you posted a link in #yelling.


nicklambourne:
Approach is a bit simple, maybe try a regex match instead? 
e.g. https://stackoverflow.com/a/3809435

paulrusthewalrus:
Whoops, didn't see your message before I made this new PR. Will try that now :)

kentonlam:
This closes #534.

uqcsbot/scripts/yelling.py
```diff
@@ -2,6 +2,7 @@
 
 from random import choice, random
 from re import sub, UNICODE
+import re
```
nicklambourne, line 5: `import re` should go first in imports because it's a global import (the whole package).

nicklambourne (changes requested):
First, have you tested this in the test workspace? Can you post some screenshots of it working? Will also need a test or two at some point.
Second, I don't think with the current setup that it will work with this one:

> f"DID YOU SAY \n>>>{mutate_minuscule(text)}".upper(),

You either need to put the link back in somehow or remove this from the available options if a link exists.

bradleysigma:
Odd things may happen if the bot chooses the `f"DID YOU SAY \n>>>{mutate_minuscule(text)}".upper()` response when there is both a link and other, non-link minuscule characters in the message.

paulrusthewalrus:
Tests for the new PR:
Screen Shot 2020-10-03 at 5 34 07 pm

nicklambourne:
Might I suggest using re.match() to identify where the links were so you can reinsert them after analysis and mutation?

paulrusthewalrus:
Screen Shot 2020-11-19 at 3 18 06 pm

nicklambourne (approved):

nicklambourne:
@kentonlam @TomCranitch Can one of you two have a look at this? I changed a little more than I expected to in the cleanup.

TomCranitch (approved):

PR #545, opened 2020-10-03
create factorial response

bradleysigma:
Calculates factorials. Aggressively.
image

### DIFF
--- a/test/test_factorial.py
+++ b/test/test_factorial.py
@@ -1,68 +1,63 @@
-from test.conftest import MockUQCSBot, TEST_CHANNEL_ID
-from unittest.mock import patch
+from test.conftest import MockUQCSBot, TEST_CHANNEL_ID, TEST_USER_ID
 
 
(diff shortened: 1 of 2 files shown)


jdcaperon:
I don't see it being triggered a lot, but people will probably complain about it.

nicklambourne:
So if I say something like "They only gave you 18!," this is going to trigger?

bradleysigma:
Correct. Because it's more amusing this way.

kentonlam (commented):
Code looks fine. I'm more worried that it'll be annoying rather than funny, particularly since it triggers automatically and on all channels.

kumar-ish:
I also don't think it'll be triggered _too_ often but it'll mildly annoying. With that said, if you make it to be a(n im)probabilistic event (say, `if random.random() < 0.1`) I think it's fine.

nicklambourne:
@bradleysigma I think Madhav's suggestion is good. If you set the probability at 5% or less and mock the probability for the test I'll review and merge this.

PR #561, opened 2021-01-15
Update emojify.py

bradleysigma:


nicklambourne (approved):
While I think it's practically impossible for people to enter the delete character into Slack I find it amusing enough to warrant keeping.

PR #565, opened 2021-03-29
Add !link command

nicklambourne:
Adds a user accessible key value store, e.g:

Screen Shot 2021-03-30 at 1 41 15 am
Screen Shot 2021-03-30 at 1 42 39 am
Screen Shot 2021-03-30 at 1 47 48 am

This has the added benefit of adding the "brain" back into the bot and provides a template for further (and easy) use of the database though sqlalchemy orm models.

- [x] Still needs tests and more thorough manual testing.
- [ ] Need to set up the UQCSBOT_DB_URI env var in the server config.


uqcsbot/models.py
```diff
@@ -0,0 +1,17 @@
...
+class Link(Base):  # type: ignore
+    __tablename__ = 'links'
+
+    id = Column("id", Integer, primary_key=True, nullable=False, autoincrement=True)
+    key = Column("key", String, nullable=False)
+    channel = Column("channel", String, nullable=True)
+    value = Column("value", String, nullable=False)
+
+    def __repr__(self):
+        return f"<Link: ({self.key}, {self.channel} = {self.value})>"
```
kentonlam, line 17: Nit: Can this be a more conventional `Link(self.key, ...)` form?

uqcsbot/scripts/link.py
```diff
@@ -0,0 +1,161 @@
...
+def set_link_value(key: str, value: str, channel: str,
+                   link_scope: LinkScope, override: bool) -> Tuple[SetResult, str]:
+    """
+    Sets a corresponding value for a particular key. Keys are set to global by default but this can
+    be overridden by passing the channel flag. Existing links can only be overridden if the
+    override flag is passed.
+    :param key: the lookup key for users to search the value by
+    :param value: the value to associate with the key
+    :param channel: the name of the channel the set operation was initiated in
+    :param link_scope: required to be True if the association is to be specific to the channel
```
kentonlam, line 39: Comment references old bool behaviour.

uqcsbot/scripts/link.py
```diff
@@ -0,0 +1,161 @@
...
+        result = SetResult.NEW_LINK_SUCCESS
+    session.add(Link(key=key, channel=link_channel, value=value))
+    session.commit()
+    session.close()
+    return result, value
+
+
+def get_link_value(key: str,
+                   channel: str,
+                   link_scope: LinkScope) -> Tuple[Optional[str], Optional[str]]:
```
kentonlam, line 63: link_scope should be Optional?

uqcsbot/scripts/link.py
```diff
@@ -0,0 +1,161 @@
...
+                                                        color=Color.YELLOW)._resolve()])
+
+    channel = bot.channels.get(command.channel_id)
+    if not channel:
+        return bot.post_message(command.channel_id, "", attachments=[
+            Attachment(SectionBlock("Cannot find channel name, please try again."),
+                       color=Color.YELLOW)._resolve()
+        ])
+    else:
+        channel_name = bot.channels.get(command.channel_id).name
```
kentonlam, line 131: Don't need else block because of return and can just use channel.name instead.

uqcsbot/scripts/link.py
```diff
@@ -0,0 +1,161 @@
...
+
+    # Retrieve a link
+    if not args.value:
+        link_value, source = get_link_value(key=args.key,
+                                            channel=channel_name,
+                                            link_scope=link_scope)
+        if_channel_flag = f" in channel `{channel_name}`" if args.channel_flag else ""
+        response = f"{args.key} ({source if source == 'global' else channel_name}): " \
+                   f"{link_value}" if link_value else \
+                   f"No link found for key: `{args.key}`{if_channel_flag}"
```
kentonlam, line 144: A bit of a mess with all the ternaries 😅 Can we split it into something like this instead?
```python
channel_text = f" in channel `{channel_name}`" if args.channel_flag else ""
source_text = source if source == 'global' else channel_name
result_text = link_value if link_value else
    f"No link found for key: `{args.key}`" + channel_text
response = f"{args.key} ({source_text}): {result_text}"
```

uqcsbot/scripts/link.py
```diff
@@ -0,0 +1,161 @@
...
+
+    # Set a link
+    if args.key and args.value:
+        result, current_value = set_link_value(key=args.key,
+                                               channel=channel_name,
+                                               value=" ".join(args.value),
+                                               link_scope=link_scope,
+                                               override=args.override)
+        color = Color.YELLOW if result == SetResult.NEEDS_OVERRIDE else Color.GREEN
+        response = f"{args.key} ({channel_name if args.channel_flag else 'global'}): " \
```
kentonlam, line 158: Not as bad as above, but can we pull the ternary into its own variable?

kentonlam (changes requested):
Looks pretty good and will surely be useful in future. The message formatting is a bit arcane and should be cleaned up, but apart from that looks good. 

Do we need the empty test/test_utils.py file?

kentonlam (approved):
Looks good, thanks for cleaning it up. What should UQCSBOT_DB_URI be on the server?

JamesDearlove (approved):
Looks like a pretty useful feature, I approve

PR #567, opened 2021-04-20
Add psycopg2 to setup.py requirements.

kentonlam:
This fix works on my machine. Hopefully fixes prod problems from #565.


JamesDearlove (approved):
What could possibly go wrong

PR #568, opened 2021-09-11
Updated to use Poetry, Dockerised and setup GitHub actions

JamesDearlove:
The great migration away from uqcs1, part 2.

This should just work tm


nicklambourne:
Merging without review:
shame-gameofthrones

